PR #3, opened 2015-01-07
Better handling of relative months and years

mruwek:
- Increment months strictly: 01.01.2015 + 1m = 01.02.2015.
- If there are fewer days in target month, date is rounded down:
  31.10.2015 + 1m = 30.11.2015.
- Years are incremented as 12 months.
- Leap years are 366 days long: 01.03.2015 + 1y = 01.03.2016 (not
  29.02.2016 as it would have been if year was equal to 365d)

### DIFF
--- a/test/RelativeDateTest.py
+++ b/test/RelativeDateTest.py
@@ -43,12 +43,40 @@ class RelativeDateTester(TopydoTest.TopydoTest):
         self.assertEquals(result, date.today() + timedelta(weeks=1))
 
     def test_one_month(self):
-        result = relative_date_to_date('1m')
-        self.assertEquals(result, date.today() + timedelta(30))
+        test_date = date(2015, 1, 10)
+        result = relative_date_to_date('1m', test_date)
+        self.assertEquals(result, date(2015, 2, 10))
+
+    def test_one_month_ext(self):
+        test_date1 = date(2015, 1, 29)
+        test_date2 = date(2016, 1, 31)
+        test_date3 = date(2015, 12, 31)
+        test_date4 = date(2015, 7, 31)
+        test_date5 = date(2015, 10, 31)
+
+        result1 = relative_date_to_date('1m', test_date1)
+        result2 = relative_date_to_date('1m', test_date2)
+        result3 = relative_date_to_date('1m', test_date3)
+        result4 = relative_date_to_date('1m', test_date4)
+        result5 = relative_date_to_date('1m', test_date5)
+
+        self.assertEquals(result1, date(2015, 2, 28))
+        self.assertEquals(result2, date(2016, 2, 29))
+        self.assertEquals(result3, date(2016, 1, 31))
+        self.assertEquals(result4, date(2015, 8, 31))
+        self.assertEquals(result5, date(2015, 11, 30))
 
     def test_one_year(self):
-        result = relative_date_to_date('1y')
-        self.assertEquals(result, date.today() + timedelta(365))
+        test_date = date(2015, 1, 10)
+        result = relative_date_to_date('1y', test_date)
+        self.assertEquals(result, date(2016, 1, 10))
+
+    def test_leap_year(self):
+        test_date = date(2016, 2, 29)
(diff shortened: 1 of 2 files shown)


bram85:
Thanks for your contribution!
Can you check whether the testcases are still valid?
Run `python setup.py test` to check (this command only applies to the master branch at the moment).

mruwek:
> Can you check whether the testcases are still valid?

I've updated the testcases to reflect changes in meaning of 1y and 1m and I've tried them with my system-date set to: 2016-02-29, 2015-01-31 and 2015-08-01. Each time test_one_month and test_one_year passed ok. Surprisingly though, test_do_subtasks2 failed if topydo.egg-info dir is already present (it passes ok if I run `rm topydo.egg-info -r` before `python setup.py test`):

```
======================================================================
FAIL: test_do_subtasks2 (test.DoCommandTest.DoCommandTest)
----------------------------------------------------------------------
  File "/home/jacek/dev/topydo/test/DoCommandTest.py", line 86, in test_do_subtasks2
    self.assertEquals(self.output, result)
AssertionError: '|  3| Baz p:1\n|  2| Bar p:1\nCompleted: x 2015-01-08 Foo id:1\n' != '|  2| Bar p:1\n|  3| Baz p:1\nCompleted: x 2015-01-08 Foo id:1\n'

----------------------------------------------------------------------
Ran 375 tests in 1.180s

FAILED (failures=1)
```

Such problem doesn't exist in master branch. Every iteration of `python setup.py test` passes ok. I have no idea what is causing this difference in output sorting.

bram85:
I see test_do_subtasks2 fail as well when testing an unrelated change. So it's a safe bet that you didn't break that test. Have to investigate where this comes from, though...

mruwek:
> I see test_do_subtasks2 fail as well when testing an unrelated change. (...) Have to investigate where this comes from, though...

If it can help, I've found that, deleting content of `topydo.egg-info/entry_points.txt` before running the tests makes test_do_subtasks2 pass ok. So it seems related with that particular file.

> So it's a safe bet that you didn't break that test.

Waiting then for your code review :)

mruwek:
> The leap day should be tested unconditionally, regardless of the current system time.

Agreed, but how do we want to implement that? Test against fixed leap day? For instance:

```python
def test_leap_year(self):
    test_date = date(2016, 2, 28)
    result = relative_date_to_date('1y', test_date)
    self.assertEquals(result, test_date + timedelta(366))
```

> I'm not so happy with the conditionals in this test and the test below.

If we want to test for one month from random date (`date.today(()`) we're going to need them, because "one month" is kind of complex idea. I see now that there should be even one more conditional for Dec 31st in test_one_month. It is then either "conditionals way" or "fixed-test-dates" way.

mruwek:
> The leap day should be tested unconditionally, regardless of the current system time.

Agreed, but how do we want to implement that? Test against fixed leap day? For instance:

```python
def test_leap_year(self):
    test_date = date(2016, 2, 28)
    result = relative_date_to_date('1y', test_date)
    self.assertEquals(result, test_date + timedelta(366))
```

> I'm not so happy with the conditionals in this test and the test below.

If we really want to test for one month from random date (`date.today(()`) we're going to need them, because "one month" is kind of complex idea. I see now that there should be even more conditionals in test_one_month (on Dec 31st result will be ok, but test will fail). So it's either "conditionals way" or "fixed-test-dates way". Personally I prefer second one and I will push soon another change with bunch of tests based on http://github.com/dateutil/dateutil approach.

bram85:
Yes, the second approach is OK. If the leap day test works for 2016, I'm quite confident it will also work for other leap years. And yes, also include Dec 31st because I was already wondering about that while looking at the code.
Thank you.

mruwek:
> And yes, also include Dec 31st

It's already there - take a look at `test_date 3` in `test_one_month_ext()`  in commit: da473a62c439c09551b47032a24e7915bfcec35c

bram85:
Ah, looks good. Somehow I missed the fact that you made a commit.

PR #5, opened 2015-02-07
Allow multiple tasks to be prioritized at once

mruwek:
I mean something like:

`topydo pri 1 2 3 4 5 6 7 8 9 10 B`


bram85:
Thank you.

PR #7, opened 2015-02-24
Different colors for context and projects

kidpixo:
I modified PrettyPrinter.py to have different colors for context and projects.
I added the `CONTEXT_COLOR` variable and a lambda function to the re.sub substitution to check if the patterns begins with `+` of `@` and attach the relevant color.

Would be nicer to add the color definition in the configuration file.

Cheers.


kidpixo:
I added also a different color for Metadata (defined in the gina's `key:value` format [here](https://github.com/ginatrapani/todo.txt-cli/wiki/The-Todo.txt-Format)) and for links.

bram85:
Hi, thanks for your patch.
I'd like to ask you to reapply the change you did to the pp_color function in the master branch instead of the stable branch. The stable branch is for bugfixes only.
Unfortunately, there are some major differences regarding pretty printing in master. The coloring code has been moved to PrettyPrinterFilter.py, I think you can copy-paste your change into that file.
I could've done that myself, but then we'd lose your name in the Git history.

Also make sure that the tests are still OK:

    python setup.py test

And I agree with your comment that the colors should be in a configuration file, but that's a bit of work (unless we want to have verbatim ASCII escape codes in the config file, not really user-friendly).

kidpixo:
I'm closing this pull request and open a new one for `master`. I did trash the history a little, it was my first attempt with `git revert`.

PR #8, opened 2015-02-28
Give 'postpone' command possibility to work with multiple todos

mruwek:
I took implementation from previous analogous effort with 'pri', wrote some new tests and corrected synopsis & help. Waiting for review :)

### DIFF
--- a/test/PostponeCommandTest.py
+++ b/test/PostponeCommandTest.py
@@ -164,6 +164,53 @@ class PostponeCommandTest(CommandTest.CommandTest):
         self.assertEquals(self.output, "|  1| Foo due:{}\n".format(due.isoformat()))
         self.assertEquals(self.errors, "")
 
+    def test_postpone14(self):
+        command = PostponeCommand(["1", "2", "1w"], self.todolist, self.out, self.error)
+        command.execute()
+
+        due = self.today + timedelta(7)
+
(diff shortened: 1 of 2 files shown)


mruwek:
> Perhaps it would be useful to print which todo ID was invalid

Ok, so here are 2 alternatives:

1. Even if only one of many IDs was invalid, don't do anything and print something like:
```
"Foo": Invalid todo number given.
"999": Invalid todo number given.
```

2. Proceed with correct ones and print in same manner as above invalid ones.

> Note that some test cases will break if you touch this string.

I'm aware of that and as soon as we decide which alternative to choose, I'll start implementing it and reviewing the tests.

bram85:
I think option 1 is better, the exit status of the script would be 0 if everything was successful, and 1 otherwise. Just in case someone checks the exit status.

mruwek:
> I think option 1 is better

If You are okay with code pushed above, I will start reviewing the tests.

bram85:
Looks good to me.

Overall, the function has become a bit long and deeply nested (for my personal style). That's no blocker at all, if you fix up the tests I'll merge your code. But I guess it's subject to a little bit of refactoring.

mruwek:
I've edited error message to maintain old behavior if only one todo ID was supplied on the command-line, wrote one new test and squashed my commits.

PR #9, opened 2015-03-02
Adding colors to Context,Projects,Metadata and Link.

kidpixo:
I reverted the change in `stable` and pushed to `master` for merge.
The original commits in `stable` are listed here for reference:
- added Metadata and Link colors to PrettyPrinter.py 0d18df88c074122fffeacb4c234100323dfd62d0
- modified PrettyPrinter.py to have different colors for context and projects 0db21ed53eae6eaa4a3f7e7dd57630a76d1d04e6


bram85:
Applying your patch causes test cases to fail, I'm not sure why. It has something to do with functionality to hide tags altogether (see the corresponding filter in PrettyPrinterFilter.py), somehow it doesn't play well with the tag coloring in your patch. Have to look at it later.

bram85:
I had another look at it and I'm afraid I cannot merge it in at this point.

I see ANSI garbage in the output when using an URL tag, part of the color code for METADATA ends up in the output, using a task like this:

`(C) Foo url:http://example.org`

I see this:

`(C) Foo 1;32murl:http://example.org`

If you find the cause of this bug, it would be nice if you could add a test to cover this so it won't happen again.

About the failing tests, most of the tests (not all) can be fixed with a patch on the `View` class, by adding the color filter after the for loop in the `pretty_print` function. There's still one test case that fails after this change. I'll see if I can commit this change to the master branch.

bram85:
Pushed a fix to master, if you rebase you can get rid of the failing tests. Still the ANSI garbage should be looked at.

kidpixo:
Hi, 
I looked at the garbage, it is due to my assumption that an url will be alone in the text and not as part of metadata pair.
That's why I have a color for general metadata and one for link:

    METADATA_COLOR = '\033[1;32m' # color for @ metadata : green
    LINK_COLOR     = '\033[4;36m' # color for @ links : cyan/underline

Your example fits in both categories, my code applies:
- first   the `METADATA_COLOR` to `url:http://example.org` 
- then `LINK_COLOR` to the `http://example.org` effectively breaking the ASCII escape sequence.

The original idea was to visualize links as more or less as html hyperlink (blue-underlined).

My proposal is to change my code to check if a link is part of a metadata `key:value` pair and skip it in case.
Do you agree with that?
Cheers.

bram85:
I found out about it because I personally use an `url` tag for things like that, yes.
If I understood you correctly, you will skip the `LINK_COLOR` when it's part of a tag? So `url:http://example.org` will get the `METADATA_COLOR`. That's OK with me.

kidpixo:
Nice,
I'm on it.

Thanks for your work!



---
"The Answer to the Great Question …"
"Of Life, the Universe and Everything ..." said Deep Thought.
"Is ... " said Deep Thought, and paused.
"Forty-two," said Deep Thought, with infinite majesty and calm.
[Douglas A., The Hitchhiker's Guide to the Galaxy]

On Tue, May 5, 2015 at 10:23 AM, Bram Schoenmakers <notifications@github.com

kidpixo:
I just discover gitter.im, why don't you add a chat room there?

kidpixo:
Now should be ok.
I add a check to change multiple occurrence of `\033[0m` just for clarity.

bram85:
I don't see any change in this PR. Forgot to push something?

kidpixo:
I did push to stable, my bad. Let me rebase and push to master.

---
"The Answer to the Great Question …"
"Of Life, the Universe and Everything ..." said Deep Thought.
"Is ... " said Deep Thought, and paused.
"Forty-two," said Deep Thought, with infinite majesty and calm.
[Douglas A., The Hitchhiker's Guide to the Galaxy]

kidpixo:
Now should be ok. This rebase/rest thing is a nightmare every time!.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -32,13 +32,16 @@ def filter(self, p_todo_str, _):
...
+    'A': '\033[0;36m', # cyan
+    'B': '\033[0;33m', # yellow
+    'C': '\033[0;34m'  # blue
 }
 
-PROJECT_COLOR = '\033[31m' # red
-NEUTRAL_COLOR = '\033[0m'
+PROJECT_COLOR  = '\033[1;31m' # color for + keyword  : red
+CONTEXT_COLOR  = '\033[1;35m' # color for @ keyword  : magenta
+METADATA_COLOR = '\033[1;32m' # color for @ metadata : green
```
mruwek: Drop ' @' from ' @ metadata' in comment as metadata doesn't start with '@'. Same for '@ link'

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -57,15 +60,21 @@ def filter(self, p_todo_str, p_todo):
             except KeyError:
                 pass
 
-            p_todo_str = color + p_todo_str + NEUTRAL_COLOR
-
+            p_todo_str = '%s%s%s' % (color, p_todo_str, NEUTRAL_COLOR)
```
mruwek: What's the difference between this line and original one? Correct me if I'm wrong, but both should give same result.

mruwek:
> This rebase/rest thing is a nightmare every time!

Rebase is actually pretty cool. Consider `--interactive` mode and next time think about creating separate branch for pull-request.

topydo/lib/PrettyPrinterFilter.py
kidpixo: You are right, here is purely cosmetic. 
Although I use only Python 2.x, I believe the second form is also Python 3.x compliant.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -57,15 +60,21 @@ def filter(self, p_todo_str, p_todo):
...
             if config().highlight_projects_contexts():
                 p_todo_str = re.sub(
                     r'\B(\+|@)(\S*\w)',
-                    PROJECT_COLOR + r'\g<0>' + color,
+                    #PROJECT_COLOR + r'\g<0>' + color,
```
mruwek: If You don't use that line maybe delete it, instead of commenting out?

mruwek:
@kidpixo
> Now should be ok.

Works good.

> This rebase/rest thing is a nightmare every time!.

Rebase is actually pretty cool. Consider `--interactive` mode and think about creating separate branches for pull requests (i.e 'colors-in-metadata-and-links').

topydo/lib/PrettyPrinterFilter.py
kidpixo: Vestiges of old stuff. I'm ok with deleting it.

kidpixo:
@mruwek  thanks, that's a good idea. I just found [Explain Git with D3](http://onlywei.github.io/explain-git-with-d3/) to get it visually, now is clear.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -57,15 +60,21 @@ def filter(self, p_todo_str, p_todo):
...
+                    lambda m: (
+                        CONTEXT_COLOR if m.group(0)[0] == "+"
+                        else PROJECT_COLOR)+m.group(0)+color,
                     p_todo_str)
-
-            p_todo_str += NEUTRAL_COLOR
+            p_todo_str = re.sub(r'\w(\S*:[^/\s]\S*)\w',METADATA_COLOR+r'\g<0>'+color,p_todo_str)
+            # add LINK_COLOR to anything like :  (one group od charater and a :)(// and non whitespace chars)
```
bram85: Please revise the comment, it doesn't really make it clearer to me (or my future self in 1 year). Maybe just simplify it to something like: add LINK_COLOR to any word that resembles an URL with :// in it

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -57,15 +60,21 @@ def filter(self, p_todo_str, p_todo):
...
                     r'\B(\+|@)(\S*\w)',
-                    PROJECT_COLOR + r'\g<0>' + color,
+                    #PROJECT_COLOR + r'\g<0>' + color,
+                    lambda m: (
+                        CONTEXT_COLOR if m.group(0)[0] == "+"
+                        else PROJECT_COLOR)+m.group(0)+color,
                     p_todo_str)
-
-            p_todo_str += NEUTRAL_COLOR
+            p_todo_str = re.sub(r'\w(\S*:[^/\s]\S*)\w',METADATA_COLOR+r'\g<0>'+color,p_todo_str)
```
bram85: Why the `\w` characters outside the regex group? Why not `(\w+:[^/\s]\w+)`  ?

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -57,15 +60,21 @@ def filter(self, p_todo_str, p_todo):
...
+            # add LINK_COLOR to anything like :  (one group od charater and a :)(// and non whitespace chars)
+            p_todo_str = re.sub(r'\s(\w*:){1}(//\S*\w)',' '+LINK_COLOR+r'\1\2'+color,p_todo_str)
```
bram85: Same question here about `\w` outside the group.

topydo/lib/PrettyPrinterFilter.py
kidpixo: You;re right, but it's essentially the same here, due to the `\g<0>` backreference.
From [7.2. re — Regular expression operations — Python 2.7.10rc0 documentation](https://docs.python.org/2/library/re.html) : 

    The backreference \g<0> substitutes in the entire substring matched by the RE.

topydo/lib/PrettyPrinterFilter.py
kidpixo: how sounds: 
   add LINK_COLOR to any word that resembles an URL with `//` and max 1 repetition of `WORD:` before 

because the whole point to differentiate between `key:value` and `key:http://url` is that we have one group before the `:`, see dummy text 

    Lorem Ipsum Foo url:http://example.org http://example.org

kidpixo: Example of dummy text with different priority with my color scheme.
 
The only but I catch it that `key:value` at the beginning of the line will not be catch, but I guess it is not common.

topydo/lib/PrettyPrinterFilter.py
kidpixo: see answer to your note [r29763811](https://github.com/bram85/topydo/pull/9#discussion_r29771538) above.

topydo/lib/PrettyPrinterFilter.py
bram85: Please rewrite the expression.

My (untested) suggestion would be `\b\S+:[^/\s]\S+\b`, the `\b` takes care of the case where the tag is at the start or end of the line. Unlikely, but it's no additional effort to cover.

topydo/lib/PrettyPrinterFilter.py
bram85: Perhaps rewrite to `\b\S+://\S+\b` ? This matches the case when a word contains `://`. Or did I miss a corner case?

topydo/lib/PrettyPrinterFilter.py
kidpixo: That works great, I check regex using this tool [Online regex tester and debugger: JavaScript, Python, PHP, and PCRE](https://regex101.com/#python) and it cover al cases.

topydo/lib/PrettyPrinterFilter.py
kidpixo: Using the tool as in [pull/9#discussion_r29783803](https://github.com/bram85/topydo/pull/9#discussion_r29783803) you can see that this option match both the combination of `key:value` and `key:http://url`. 
In the second keys you can split it as `key` and value as `http://url`. 
Something like : 

    [^:]\b(\w+?)://\S+\b

works, but again it doesn't cover beginning of the line URL, but is a corner case.

kidpixo: [Pythex: a Python regular expression editor](http://pythex.org/?regex=%5B%5E%3A%5D%5Cb(%5Cw%2B%3F)%3A%2F%2F%5CS%2B%5Cb&test_string=http%3A%2F%2Fexample.org%20http%3A%2F%2Fexample.org%20Lorem%20Ipsum%20Foo%20url%3Ahttp%3A%2F%2Fexample.org%20http%3A%2F%2Fexample.org&ignorecase=0&multiline=0&dotall=0&verbose=0) is even better, I can link you the test directly.

kidpixo: Got it, it was 

    (^|\s)(\w*:)(//\S*\w)

found obviously on stackoverflow [python - regex line start in character set - Stack Overflow](http://stackoverflow.com/a/13199255/1435167) . Test is here on [Pythex](http://pythex.org/?regex=(%5E%7C%5Cs)(%5Cw*%3A)%7B1%7D(%2F%2F%5CS*%5Cw)&test_string=http%3A%2F%2Fexample.org%20http%3A%2F%2Fexample.org%20Lorem%20Ipsum%20Foo%20url%3Ahttp%3A%2F%2Fexample.org%20http%3A%2F%2Fexample.org&ignorecase=0&multiline=0&dotall=0&verbose=0).

kidpixo: That should cover all the cases I think. 
screen shot 2015-05-06 at 20 19 55

bram85: Thanks for the Pythex links. I would suggest the following:

`(^|\s)(\w+:){1}(//\S+)` ([link](http://pythex.org/?regex=%28^|\s%29%28\w%2B%3A%29{1}%28%2F%2F\S%2B%29&test_string=http%3A%2F%2Fexample.org%20http%3A%2F%2Fexample.org%20Lorem%20Ipsum%20Foo%20url%3Ahttp%3A%2F%2Fexample.org%20http%3A%2F%2Fexample.org%20%3A%2F%2Fexample.org&ignorecase=0&multiline=0&dotall=0&verbose=0))

Because your expression would match `://example.org`.

kidpixo: agreed, well played.

topydo/lib/PrettyPrinterFilter.py
kidpixo: what about the comment?

topydo/lib/PrettyPrinterFilter.py
mruwek: > I believe the second form is also Python 3.x compliant.

Concatenating strings is still valid operation on python 3.x Consider https://docs.python.org/3/library/stdtypes.html#common-sequence-operations

So it's up to @bram85 to decide which option to choose.

bram85: And here I was thinking that `str.format()` is preferred over the `%` operator. In the past I actually replaced `%` string compositions with `.format()`. So I'd rather see either the concatenation, or use `"{}{}{}".format(...)`.

topydo/lib/PrettyPrinterFilter.py
bram85: Describe the cases you want to distinguish:
* key:value
* url:http://example.org
* http://example.org

mruwek: Wouldn't something like:

>  add LINK_COLOR to any valid URL specified outside of the tag

be sufficient?

mruwek:
@kidpixo 
Any progress on this? I've pushed recently to my [own fork](https://github.com/mruwek/topydo/commits/colors) commit with corrections in comments and updated versions of regexps discussed above. If you're happy with such changes, feel free to cherry-pick 9793690c29c822c823cbff4e5295e838090e2b5f from there and update this pull request.

If I can suggest anything I would also review commit-message of e6722a2ac72ecd0bcf8df3cc134820c123c68199, so it would contain information about added functionality rather than about reverting from stable branch and pushing to master.

kidpixo:
hi @mruwek I'm fine with you branch. When I have time I'll merge in my repo.
Thanks!

mruwek:
@kidpixo
I don't know if merging branch with only one distinct commit is the best solution in this case. Maybe try to [cherry-pick](http://git-scm.com/docs/git-cherry-pick) it:
```shell
git remote add mruwek http://github.com/mruwek/topydo
git fetch mruwek
git checkout master
git cherry-pick 9793690c29c822c823cbff4e5295e838090e2b5f
```

kidpixo:
@mruwek I got your [9793690](https://github.com/bram85/topydo/commit/9793690c29c822c823cbff4e5295e838090e2b5f) in my master. Thanks  for the tip.

bram85:
Code looks OK, but can you squash commits e6722a2 and 5a7943a, and enter a proper commit message please?

mruwek:
@kidpixo
I'm glad I could help. You can squash your commits with [interactive rebase](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits): `git rebase -i 2a9be13f`. It will open your default editor where you can change "pick" to "squash" on the line of 5a7943a and after that editor for new "squashed" commit message will pop out, so you can enter proper informations there.

bram85:
Cherry-picked the relevant commits into master. Thank you for your contribution.

Issue #2, opened 2015-01-06
Provide "edit" command


mruwek:
It would be nice to have possibility to edit todo.txt in $EDITOR. I can imagine 3 ways of using this command:

1. `todo edit` - open todo.txt in $EDITOR
2. `todo edit archive` - open done.txt
3. `todo edit <NUMBER>` - open todo.txt and go to specified task/line

Bottom line: inspired by work of Dave Hein @dhein
http://github.com/dhein/todo.txt-cli/blob/pub-addons-new_export2/edit

bram85:
Didn't know that extension by Dave Hein, but I was actually already working on this.
In a local branch that, I didn't push yet, I already have an 'edit' subcommand that implements item 1 in your list.

I had a different implementation of "edit <NUMBER>" in mind, being to open a temporary file with that specified todo item. You could be able to specify more numbers, or maybe even a +Project to edit all items in that project. That last item is not implemented yet.

Regarding item 2, I'd say to use a flag (-d) instead a full word. "archive" could also be a regular expression that matches items with the word archive in it.

mruwek:
> I was actually already working on this.

Great :) I'm waiting then for the update.

> You could be able to specify more numbers, or maybe even a +Project to edit all items in that project.

That would be very useful. Any ETA on this?

> Regarding item 2, I'd say to use a flag (-d) instead a full word.

Agreed.

bram85:
> > You could be able to specify more numbers, or maybe even a +Project to edit all items in that project.

> That would be very useful. Any ETA on this?

ETA is difficult to tell. I could deliver functionality 1) very soon to master because that's basically ready. I just waited because I assumed to be the only user :)

At the moment I'm a bit busy with other things, but item 3) is something I miss myself in my day-to-day work so probably I'll scratch that itch somewhere this month or next month.

kidpixo:
+1 on this!

mruwek:
I really believe that code above is not ready for merge, but hey, it works!

Waiting for insights, comments and patches :)

>  or maybe even a +Project to edit all items in that project

For that to happen we need to implement it in `TodoListBase.todo` function. `pri`, `postpone`, `do`, `del` would also benefit from such extension.

bram85:
I'll leave this issue open since the `archive` option is not implemented (use case 2 in your original report). It's up to you what to do: close this issue, now the `edit` subcommand is merged, or leave it open in case you'd like to implement the archive case.

mruwek:
[...]

PR #11, opened 2015-03-27
Very early implementation of editing todos in $EDITOR

mruwek:
implements #2

- `topydo edit` will still open todofile in $EDITOR
- `topydo edit 1 2 3` will open tmpfile with corresponding todos

### DIFF
--- a/topydo/lib/EditCommand.py
+++ b/topydo/lib/EditCommand.py
@@ -133,11 +133,24 @@ class EditCommand(MultiCommand, ListCommand):
                         'Check your $EDITOR and/or $PATH')
 
     def usage(self):
-        return """Synopsis: edit"""
+        return """Synopsis:
+  edit
+  edit <NUMBER1> [<NUMBER2> ...]
+  edit -e [-x] [expression]"""
 
     def help(self):
-        return """Launches a text editor with the todo.txt file.
+        return """\
+Launches a text editor to edit todos.
+
+Without any arguments it will just open todo.txt file. Alternatively it can
+edit todo item(s) with the given number(s) or edit relevant todos matching
+given expression. Definition of relevant todo is given in the help message
+of `ls` command.
 
 By default it will use $EDITOR in your environment, otherwise it will fall back
 to 'vi'.
+
+-e : Treat the subsequent arguments as an expression.
+-x : Edit *all* todos matching the expression (i.e. do not filter on
+     dependencies or relevance).
 """
--- a/topydo/lib/EditCommand.py
+++ b/topydo/lib/EditCommand.py
@@ -15,27 +15,122 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import os
-from subprocess import call
+from subprocess import call, check_call, CalledProcessError
+import tempfile
 
-from topydo.lib.Command import Command
+from topydo.lib.ListCommand import ListCommand
+from topydo.lib.MultiCommand import MultiCommand
 from topydo.lib.Config import config
+from topydo.lib.Todo import Todo
+from topydo.lib.TodoListBase import InvalidTodoException
+from topydo.lib.TodoList import TodoList
+from topydo.lib.PrettyPrinterFilter import PrettyPrinterNumbers
 
-class EditCommand(Command):
+class EditCommand(MultiCommand, ListCommand):
     def __init__(self, p_args, p_todolist, p_output, p_error, p_input):
         super(EditCommand, self).__init__(p_args, p_todolist, p_output,
             p_error, p_input)
 
+        self.is_expression = False
+
+    def _process_flags(self):
+        opts, args = self.getopt('xe')
+
+        for opt, value in opts:
+            if opt == '-x':
+                self.show_all = True
+            elif opt == '-e':
+                self.is_expression = True
+
+        self.args = args
+
+    def _todos_to_temp(self):
+        f = tempfile.NamedTemporaryFile()
+        for todo in self.todos:
+            f.write("%s\n" % todo.__str__())
+        f.seek(0)
+
+        return f
+
+    def _todos_from_temp(self, temp_file):
+        temp_file.seek(0)
+        todos = temp_file.read().splitlines()
+
+        todo_objs = []
+        for todo in todos:
+            todo_objs.append(Todo(todo))
+
+        return todo_objs
+
+    def _open_in_editor(self, temp_file, editor):
+        try:
+            return check_call([editor, temp_file.name])
+        except(CalledProcessError):
+            self.error('Something went wrong in the editor...')
+            return 1
+
+    def _catch_todo_errors(self):
+        errors = []
+
+        if len(self.invalid_numbers) > 1 or len(self.invalid_numbers) > 0 and len(self.todos) > 0:
+            for number in self.invalid_numbers:
+                errors.append("Invalid todo number given: {}.".format(number))
+        elif len(self.invalid_numbers) == 1 and len(self.todos) == 0:
+            errors.append("Invalid todo number given.")
+
+        if len(errors) > 0:
+            return errors
+        else:
+            return None
+
     def execute(self):
-        if not super(EditCommand, self).execute():
+        if not super(ListCommand, self).execute():
             return False
 
+        self.printer.add_filter(PrettyPrinterNumbers(self.todolist))
         try:
             editor = os.environ['EDITOR'] or 'vi'
         except(KeyError):
             editor =  'vi'
-        todo = config().todotxt()
 
-        return call([editor, todo]) == 0
+        try:
+            if len(self.args) < 1:
+                todo = config().todotxt()
+
+                return call([editor, todo]) == 0
+            else:
+                self._process_flags()
+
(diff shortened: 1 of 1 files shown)


mruwek:
Squashed 8c73a362299a2d910cb6078510182c26e767fca7 + 011fc4ecd5fd094477c0942622a15383400420c3 and rebased.

topydo/lib/EditCommand.py
```diff
@@ -15,27 +15,87 @@
...
+        todos = temp_file.read().splitlines()
+
+        todo_objs = []
+        for todo in todos:
+            todo_objs.append(Todo(todo))
+
+        return todo_objs
+
+    def _open_in_editor(self, temp_file, editor):
+        return check_call([editor, temp_file.name])
```
bram85: `check_call` may throw CalledProcessError, you should catch it somewhere.

bram85:
I think the approach looks fine, but I would have taken another approach for the user interface. Rather than enumerating the applicable IDs, I'd take an expression like the `topydo ls` subcommand. So with a `topydo edit @Home due:today` I'd get all todos matching the expression. I think that's more powerful, and the current ListCommand already knows how to do this. I think they can share this functionality, what do you think?
Note however, that enumerating the separate IDs is still a valid use case I think. Not sure how to properly combine it with the approach mentioned above at this point.

mruwek:
> (...) I think that's more powerful, and the current ListCommand already knows how to do this. I think they can share this functionality, what do you think?

You're right. I'll look into ListCommand and see how to borrow this functionality from it.

 > Note however, that enumerating the separate IDs is still a valid use case I think. Not sure how to properly combine it with the approach mentioned above at this point.

I don't like "--" flags, but I believe it's the only way. We have to decide what EditCommand should do by default and then cover the other way with some flag. For instance: `todo edit 1 2 3` would open todos with corresponding IDs (1, 2, 3), but `todo edit -e @Home due:today` would open todos matching expression (note `-e`).

mruwek:
Ok, so I have working implementation on this with mentioned `-e` approach

mruwek:
Ok, so I have working implementation with mentioned `-e` approach on my local copy, but I would like to wait for #15 to be merged before updating this one.

mruwek:
So this is it. It probably need some doctrings here and there and updated synopsis for sure, but it works in both mentioned "modes" and `topydo edit help` is still working.

Waiting for code review.

mruwek:
@bram85:
> check_call may throw CalledProcessError, you should catch it somewhere.

Done. I also catch OSError, just in case $EDITOR and/or $PATH are not set correctly.

topydo/lib/EditCommand.py
```diff
@@ -15,27 +15,121 @@
...
+                    self.get_todos(self.args)
+
+                todo_errors = self._catch_todo_errors()
+
+                if not todo_errors:
+                    temp_todos = self._todos_to_temp()
+
+                    if not self._open_in_editor(temp_todos, editor):
+                        new_todos = self._todos_from_temp(temp_todos)
+                        if len(new_todos) == len(self.todos):
```
bram85: Why this restriction?

topydo/lib/EditCommand.py
```diff
@@ -15,27 +15,121 @@
...
+                            for todo in self.todos:
+                                self.todolist.delete(todo)
```
bram85: Are dependencies between todo items properly preserved? Might need a test case to check this.

mruwek:
@bram85
> Are dependencies between todo items properly preserved?

They're not. I've overlooked that problem :P. I've tested in my local branch two solutions, which are in fact one, but differ in method calling manner.

1: Add new method in `TodoList.TodoList:`
```python
 def delete_wo_deps(self, p_todo):
        super(TodoList, self).delete(p_todo)
```
and call it instead of `TodoList.TodoList.delete()`. I'm very open to eventual renaming :)

2: Modify `TodoList.TodoList.delete()`:
```diff
--- a/topydo/lib/TodoList.py
+++ b/topydo/lib/TodoList.py
@@ -88,24 +88,27 @@ class TodoList(TodoListBase):
         self._update_parent_cache()
         self.dirty = True

-    def delete(self, p_todo):
+    def delete(self, p_todo, basic=True):
         """ Deletes a todo item from the list. """
-        try:
-            number = self._todos.index(p_todo)
+        if not basic:
+            try:
+                number = self._todos.index(p_todo)

-            for child in self.children(p_todo):
-                self.remove_dependency(p_todo, child)
+                for child in self.children(p_todo):
+                    self.remove_dependency(p_todo, child)

-            for parent in self.parents(p_todo):
-                self.remove_dependency(parent, p_todo)
+                for parent in self.parents(p_todo):
+                    self.remove_dependency(parent, p_todo)

-            del self._todos[number]
-            self._update_todo_ids()
+                del self._todos[number]
+                self._update_todo_ids()

-            self.dirty = True
-        except ValueError:
-            # todo item couldn't be found, ignore
-            pass
+                self.dirty = True
+            except ValueError:
+                # todo item couldn't be found, ignore
+                pass
+        else:
+            super(TodoList, self).delete(p_todo)

     def add_dependency(self, p_from_todo, p_to_todo):
         """ Adds a dependency from task 1 to task 2. """
```
and call it with `basic=True` argument.

So the problem with disappearing dependencies seems **fixed**. However we still want cleaning of garbage dependencies, don't we? I've added `self.todolist.clean_dependencies()` after whole "transaction" to address that issue:
```diff
--- a/topydo/lib/EditCommand.py
+++ b/topydo/lib/EditCommand.py
@@ -114,11 +114,13 @@ class EditCommand(MultiCommand, ListCommand):
                         new_todos = self._todos_from_temp(temp_todos)
                         if len(new_todos) == len(self.todos):
                             for todo in self.todos:
-                                self.todolist.delete(todo)
+                                self.todolist.delete_wo_deps(todo)

                             for todo in new_todos:
                                 self.todolist.add_todo(todo)
                                 self.out(self.printer.print_todo(todo))
+
+                            self.todolist.clean_dependencies()
                         else:
                             self.error('Number of edited todos is not equal to '
                                         'number of supplied todo IDs.')
```
but surprisingly it doesn't work as I expected. I'll show it on an example:
```
2015-04-28 test1 @test id:1
2015-04-28 test2 @test p:1
```
When I edit *test1* (`topydo edit test1`) and remove 'id:1' tag in editor, 'p:1' portion in 'test2' doesn't disappear auto-magically. It does however if I edit test1 once more.

Summary: dependencies are properly **preserved**, but garbage collection is not immediate.

topydo/lib/EditCommand.py
mruwek: I wanted to prevent user from accidentally removing his todos. However if You think that it is unnecessary care, I can simply remove that condition.

mruwek:
I realized, that todolist needs to be reloaded before dependencies cleaning, so I wrote some code to provide such behavior. I have also pushed part mentioned above for preserving proper dependencies.

Seems like only synopsis and help is missing now.

mruwek:
I have updated f12b0235007f5b5f68cd545654a7ee02c0d8bdbe to a1d7fdb, to flush todolist *without* todofile unnecessary write() and read() on todofile.

bram85:
I don't think we should worry too much about cleaning dependencies, for these reasons:

* When messing with `id` and `p` tags you're basically on your own, these tags should only be managed by topydo.
* Deleting todos is not allowed, so that wouldn't cause garbage.

I'd rather see the flushing be reverted: it adds complexity and I think it's not the responsibility of the CLI interface to handle this, but rather the responsibility of the TodoList or the EditCommand.

I think there's a third alternative of deleting an item without dependencies, is to call the `delete` of  `TodoListBase` directly from `EditCommand`. So replace the call:

```diff
--- a/topydo/lib/EditCommand.py
+++ b/topydo/lib/EditCommand.py
@@ -114,7 +114,7 @@ class EditCommand(MultiCommand, ListCommand):
                         new_todos = self._todos_from_temp(temp_todos)
                         if len(new_todos) == len(self.todos):
                             for todo in self.todos:
-                                self.todolist.delete_wo_deps(todo)
+                                super(TodoList, self.todolist).delete()

                             for todo in new_todos:
                                 self.todolist.add_todo(todo)
```

I had a brainwave of hiding `id` and `p` tags in the editor and then merge them back in somehow when reading the new items. But that requires some administration and I'm not sure if it's worth the effort. Let's first focus on finishing the current (basic) functionality.

So please revert the flush, try whether my suggestion with `super()` works and you can fix up the help and synopsis. I think it will be good for merging then.

mruwek:
@bram85:
> I think it's not the responsibility of the CLI interface to handle this, but rather the responsibility of the TodoList or the EditCommand

You're right. I believe it should be implemented in TodoList, if anywhere.

> So please revert the flush

Done.

> try whether my suggestion with `super()` works

Done. It works after fixing typo - `delete()` needs an argument.

> (...) and you can fix up the help and synopsis.

Done. Take a look, and tell me, if you want to add/change something in there.

I've also squashed my commits to not pollute history too much :)

bram85:
I'll do a few minor touch-ups on the help text, but as you can see, it's merged. Thanks a lot for your work.

PR #12, opened 2015-03-28
Append parent contexts

mruwek:
`append_parent_projects` is so useful, that I wanted this kind of functionality also for contexts :)


bram85:
Makes sense, thank you. Since I don't really use contexts, it was simply an overlooked usecase.

PR #13, opened 2015-04-07
topydo edit won't work if EDITOR is unset

mruwek:
On some systems EDITOR is unset by default, so in such situation topydo
would throw KeyError exception instead of opening todo.txt in vi.


mruwek:
Previous version of this commit was: addf7fa013be900095de9e460e5d88d8b8834ac2

bram85:
Thank you for the fix.

PR #16, opened 2015-05-04
Edit the archive file with '-d' option.

mruwek:
This implements archive option from #2.

### DIFF
--- a/topydo/lib/EditCommand.py
+++ b/topydo/lib/EditCommand.py
@@ -32,12 +32,15 @@ class EditCommand(MultiCommand, ListCommand):
             p_error, p_input)
 
         self.is_expression = False
+        self.edit_archive = False
 
     def _process_flags(self):
-        opts, args = self.getopt('xe')
+        opts, args = self.getopt('xed')
 
         for opt, value in opts:
-            if opt == '-x':
+            if opt == '-d':
+                self.edit_archive = True
+            elif opt == '-x':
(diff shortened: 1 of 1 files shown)


bram85:
Looks good, works as advertised.

Still one consideration though ;) I'm worried about the performance with huge done.txt files (hundreds or thoasands of items). It's the reason why I have separate classes `TodoListBase` and `TodoList`. For dealing with archive files, I use `TodoListBase` because the dependencies - which are mostly useless in done.txt - were being processed and consumed a lot of time (building a dependency graph, calculating parent and child items, etc.).
I think the fix would be to apply the same trick as you did with `super(..).delete(todo)` in the original patch, to call `TodoListBase.addTodo()` when dealing with an archive file.

mruwek:
I can't see a performance issue here because editing archive works only in the same manner as `topydo edit` - **direct** opening file with the editor.

I purposely didn't implement functionality of editing subset of archive in tmpfile (like with `topydo edit NUMBER1 NUMBER2` or `topydo edit -e @Home`), because I don't see practical benefits of such approach. Moreover we don't have practical means to list archived tasks with their identifers. So - in my opinion -  there is nothing to be fixed right now :)

If You however want to do something like `topydo edit  -de @Home due:today`, I can try to work on that.

bram85:
Yep, I see it now, your implementation makes sense.

PR #17, opened 2015-05-04
Add tests for EditCommand

mruwek:
So I've thought about testcases for EditCommand for some time now and here is the result. I used mock library, because it's the simplest way in my opinion to bypass all `subprocess` parts of functionality. Moreover mock is part of standard library since python >= 3.3.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,6 +10,7 @@ setup(
     url = "https://github.com/bram85/topydo",
     extras_require = {
         'ical': ['icalendar'],
+        'edit-cmd-tests': ['mock'],
     },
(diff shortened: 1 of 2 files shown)


bram85:
* Please add mock to the `extra_requires` in setup.py.
* Maybe add a test case that checks than the edited file contains the same amount of todo items.

mruwek:
> Please add mock to the extra_requires in setup.py.

Done in 23eb51f

> Maybe add a test case that checks than the edited file contains the same amount of todo items.

Done in 0a066e4

PR #25, opened 2015-05-20
Fix help in prompt-mode

mruwek:
This is done with catching TypeError and calling `topydo.cli.CLIApplicationsBase.usage`


bram85:
For now OK, ideally I would like to see a specialized help text within the prompt mode. There are some specificities in there from the original CLI. Something I'll pick up.

PR #26, opened 2015-05-20
Start prompt mode via `topydo prompt`

mruwek:
Add UILoader.py as an entry_point and decide which UI to start from there.

### DIFF
--- a/topydo/cli/UILoader.py
+++ b/topydo/cli/UILoader.py
@@ -1,0 +1,34 @@
+# Topydo - A todo.txt client written in Python.
+# Copyright (C) 2014 - 2015 Bram Schoenmakers <me@bramschoenmakers.nl>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+""" Entry file for the Python todo.txt CLI. """
+
+import sys
+from topydo.cli.CLI import CLIApplication
+from topydo.cli.Prompt import PromptApplication
+
+def main():
+    """ Main entry point of the CLI. """
+    try:
+        if sys.argv[1] == 'prompt':
+            PromptApplication().run()
+        else:
+            CLIApplication().run()
+    except IndexError:
+        CLIApplication().run()
+
+if __name__ == '__main__':
+    main()
--- a/topydo/cli/UILoader.py
+++ b/topydo/cli/UILoader.py
@@ -17,13 +17,22 @@
 """ Entry file for the Python todo.txt CLI. """
 
 import sys
+import getopt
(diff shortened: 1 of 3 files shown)


topydo/cli/UILoader.py
```diff
@@ -0,0 +1,34 @@
...
+""" Entry file for the Python todo.txt CLI. """
+
+import sys
+from topydo.cli.CLI import CLIApplication
+from topydo.cli.Prompt import PromptApplication
+
+def main():
+    """ Main entry point of the CLI. """
+    try:
+        if sys.argv[1] == 'prompt':
```
bram85: I don't think `argv[1]` works when I would run `topydo -t /some/other/todo.txt prompt`.

mruwek: It doesn't. I totally forgot about such usecase. Right now I'm thinking about 2 possible alternatives:

1. Move `CLIApplicationsBase._process_flags` into `UILoader`
2. Get opts and args, but only to check `if args[0] == 'prompt'`

I chose the latter one, although I don't like running same code twice (I mean getopt part here) . Let me know what You think about that.

bram85: After I went offline I thought of something simpler: check whether `argv[-1] == 'prompt'`. So no need to use `getopt` nor the need to maintain a copy of the allowed flags.

mruwek: Nice, but I see one problem though:
`topydo add FooBar prompt`.

bram85: `len(argv) == 2 and argv[-1] == 'prompt'`

We're getting there :)

mruwek: Um, but I believe it wouldn't be any better than `sys.argv[1]` in case of `topydo -t  foobar.txt prompt`.

bram85: Yep, you're right. In that case your approach with an additional getopt would be the best, but then I'd like to see the getopt string defined as a constant somewhere (perhaps in CLIApplicationBase) so it's defined only once.

mruwek: Or maybe create some function inside `UILoader` like `_preprocess_flags` and do there something like:

```python
def _preprocess_flags():
    args = sys.argv[1:]

    if PY2:
        args = [arg.decode('utf-8') for arg in args]

    try:
        opts, args = getopt.getopt(args, "c:d:ht:v")
    except getopt.GetoptError as e:
        error(str(e))
        sys.exit(1)

    return (opts, args)
```

and in `UILoader.main` do something like this:

```python
args = _preprocess_flags()

if args[1][0] == 'prompt':
    PromptApplication(args).run()
else:
    CLIApplication(args).run()
```

Then of course `_process_flags` would operate on args passed to objects as above, instead of on `sys.argv`.

mruwek:
@bram85
I've pushed above commit with `getopt` string defined as constant in `CLIApplicationsBase`, as You suggested. In [another branch](https://github.com/bram85/topydo/compare/prompt...mruwek:prompt-preprocess-args) though, I'm presenting different approach from my own comment above.

bram85:
I gave both options some thought because they are both appealing. But I think the version in this PR with `MAIN_OPTS` is a bit better because CLI.py and Prompt.py are still standalone applications in this way. The downside is that `getopt()` is called twice. Performance-wise a non-issue, only some slight code duplication.

mruwek:
I'm OK with that. Should I do anything more in this PR or will you just merge it in its present state?

bram85:
It looks OK, I intended to fix up some docstrings myself on top of your commits. CLI, Prompt and UILoader now all state something like _Main entry point of the CLI._

PR #27, opened 2015-05-22
Add possibility to configure colorscheme of topydo

mruwek:
- Add new module `topydo.lib.Colors` and class `Colors` for handling output colors and their ANSI codes.
- Create new config options with colors for priorities, projects, contexts, metadata, and links and place them under new section called "colorscheme".

Each configurable color is a number from range 0-255 and corresponds with relevant position in [xterm color chart](http://en.wikipedia.org/wiki/File:Xterm_256color_chart.svg). Link, metadata, project, and context colors are configured as in example below:

`link_color = 1`

 Priority colors are defined in one string, where `priority:color_number` pairs are separated by commas. Number of defined priority colors and their order in configuration string are totally arbitrary. For example:

`priority_colors = A:1,B:2,C:3,Z:4,D:27`

### DIFF
--- a/topydo.conf
+++ b/topydo.conf
@@ -30,3 +30,16 @@ ignore_weekends             = 1
 append_parent_projects      = 0
 ; Add parent contexts when adding sub todo items
 append_parent_contexts      = 0
+
+[colorscheme]
+; Configure colorscheme. Accepted values are: black, [light-]red, [light-]green,
+; [light-]yellow, [light-]blue, [light-]magenta, [light-]cyan, white
+; [light-]gray, darkgray or numbers from 0 to 255. When number is specified color
+; is matched from Xterm color chart available here:
+; http://en.wikipedia.org/wiki/File:Xterm_256color_chart.svg
+
+; priority_colors = A:cyan,B:yellow,C:blue
+; project_color = red
+; context_color = magenta
+; metadata_color = green
+; link_color = light-cyan
--- a/topydo/lib/Colors.py
+++ b/topydo/lib/Colors.py
@@ -1,0 +1,122 @@
+# Topydo - A todo.txt client written in Python.
+# Copyright (C) 2014 - 2015 Bram Schoenmakers <me@bramschoenmakers.nl>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+""" This module serves for managing output colors. """
+
+from topydo.lib.Config import config
+
+NEUTRAL_COLOR  = '\033[0m'
+
+class Colors(object):
+    def __init__(self):
+        self.priority_colors = config().priority_colors()
+        self.project_color = config().project_color()
+        self.context_color = config().context_color()
+        self.metadata_color = config().metadata_color()
+        self.link_color = config().link_color()
+
+    def _int_to_ansi(self, p_int, p_decorator='normal'):
+        """
+        Returns ansi code for color based on xterm color id (0-255) and
+        decoration, where decoration can be one of: normal, bold, faint,
+        italic, or underline.
+        """
+        decoration_dict = {
+                'normal': '0',
+                'bold': '1',
+                'faint': '2',
+                'italic': '3',
+                'underline': '4'
+        }
+
+        decoration = decoration_dict[p_decorator]
+
+        try:
+            if 256 > int(p_int) >=0:
+                return '\033[{};38;5;{}m'.format(decoration, str(p_int))
+            else:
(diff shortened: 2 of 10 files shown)


topydo/lib/Colors.py
```diff
@@ -0,0 +1,70 @@
...
+                'normal': '0',
+                'bold': '1',
+                'faint': '2',
+                'italic': '3',
+                'underline': '4'
+        }
+
+        decoration = decoration_dict[decoration]
+
+        if 256 > p_int >=0:
```
bram85: Needs an `else` clause, returning the undecorated neutral color code?

topydo/lib/Colors.py
```diff
@@ -0,0 +1,70 @@
...
+
+class Colors(object):
+    def __init__(self):
+        self.priority_colors = config().priority_colors()
+        self.project_color = config().project_color()
+        self.context_color = config().context_color()
+        self.metadata_color = config().metadata_color()
+        self.link_color = config().link_color()
+
+    def _int_to_ansi(self, p_int, decoration):
```
bram85: decoration should have default value 0, also please rename to `p_decorator` to be consistent with the rest of the code base.

topydo/lib/Colors.py
mruwek: Absolutely. I thought about it, and then forgot to implement. Patch in a minute.

topydo.conf
```diff
@@ -30,3 +30,10 @@ ignore_weekends             = 1
 append_parent_projects      = 0
 ; Add parent contexts when adding sub todo items
 append_parent_contexts      = 0
+
+[colorscheme]
+; priority_colors = A:6,B:3,C:4
```
bram85: Would be better to replace the numbers with color names.
(And add a reference to a color table for all values <256).

bram85:
I think the Colors class needs some unittests. Since it deals with user input it should be robust. It should be fed several configuration files with valid and invalid values to make sure that it handles all cases correctly.

topydo/lib/Config.py
```diff
@@ -186,6 +193,49 @@ def hidden_tags(self):
         hidden_tags = self.cp.get('ls', 'hide_tags')
         return [] if hidden_tags == '' else hidden_tags.split(',')
 
+    def priority_colors(self):
+        """ Returns a dict with priorities as keys and color numbers as value. """
+        pri_colors_str = self.cp.get('colorscheme', 'priority_colors')
+
+        def _str_to_dict(p_string):
+            pri_colors_dict = dict()
+            for pri_color in pri_colors_str.split(','):
```
bram85: `pri_colors_str` should be `p_string`.

Corresponding test-case: use an empty value for `priority_colors`.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -65,12 +61,12 @@ def filter(self, p_todo_str, p_todo):
                 p_todo_str = re.sub(
                     r'\B(\+|@)(\S*\w)',
                     lambda m: (
-                        CONTEXT_COLOR if m.group(0)[0] == "+"
-                        else PROJECT_COLOR)+m.group(0)+color,
+                        context_color if m.group(0)[0] == "+"
```
bram85: `+` should be `@` (granted, this was already wrong in the current version).

topydo/lib/Config.py
mruwek: > Corresponding test-case: use an empty value for `priority_colors`

Done. However I got some afterthoughts and I don't know whether current behavior for `priority_colors` is intuitive enough. As You can see in revised tests, empty value for `priority_colors` in config will give us default results, but empty values for other `*_color` parameters will return `NEUTRAL_COLOR`. I'm thinking about unifying this, so empty `priority_colors ` would also return `NEUTRAL_COLOR` for A, B, and C. What do You think?

bram85: Looking at it again, I wonder if falling back to `NEUTRAL_COLOR` is appropriate for any color setting. If, for instance, the project color is missing, I think it's better to use the priority color of that item (if configured and applicable).

```ini
[colorscheme]
project_color =
```

```
(C) This item would be shown +blue by default
                             ^^^^^
                             white
```

Instead, I would expect `+blue` to turn up... well... blue.

Returning to your question for the priorities, that would mean that all todo items will appear neutrally colored when they upgrade topydo, since they don't have a configuration file where `priority_colors` is defined. I think we should always use sane defaults reflecting the current behavior (topydo < 0.4).

mruwek: > Instead, I would expect +blue to turn up... well... blue.

You're right. I'll try to implement it. I'm thinking about something like this:

```python
def get_project_color(self):
        if self.project_color == '':
            return ''
        else:
            return self._get_ansi(self.project_color, 'bold')
```

This way project_color wouldn't add any ansi to output, so `+project` should retain previously defined color, which should be priority_color or NEUTRAL_COLOR (if priority_color is undefined). I'll try it out and send patch in short time.

> Returning to your question for the priorities, that would mean that all todo items will appear neutrally colored when they upgrade topydo, since they don't have a configuration file where priority_colors is defined.

topydo/lib/PrettyPrinterFilter.py
mruwek: Done.

topydo.conf
mruwek: Done.

topydo/lib/Colors.py
mruwek: > The docstring below suggests to me it should be an integer.

Oh, It could be misleading indeed. I'll try to rephrase it.

Issue #28, opened 2015-05-23
Regressions in EditCommand introduced with 5562ed7


mruwek:
There are 2 issues introduced with 5562ed7de7813ee0a6a47186cae26b72d2003101:

1. `topydo edit -d` is no longer working
2. Output of `topydo edit -e expression` is doubled. At first it prints output of `topydo ls expression` and later real output of `EditCommand`.

First issue happens because with current way of dealing with `super()` inside `execute()` we are processing flags with `ListCommand._process_flags()` and `-d` option doesn't exist there.

Second one is due to the similar cause: `EditCommand.execute()` runs `ListCommand.execute()` so we get doubled output.

I have working fix for both problems in my local branch, but I want to write relevant tests before pushing it here.

PR #29, opened 2015-05-24
Fix EditCommand inheritance problems

mruwek:
Now it inherits from newly added `topydo.lib.ExpressionCommand` to acquire `_filters` and `_view`. This way we're free from problems described in #28.


bram85:
Thanks for the fix and the improved test coverage. This won't happen again :)

PR #30, opened 2015-05-25
Add todos from file with '-f' flag.

mruwek:
Sometimes I don't have access to my todo.txt file, but I do want to save some todo items. In such cases I'm writing those items into some .txt file and later manually copy-paste them into proper todo.txt. However this way I can't benefit from topydo awesomeness (relative dates, dependencies etc.). This PR solves this issue. `AddCommand` will open file specified after '-f' flag and treat its every line as separate todo item, pre- and postprocessing it like it's done currently when adding todo item from cli.

This way we can put into foobar.txt:

```
test1 due:tom t:today @test
test2 due:fri @test before:test1
```

And topydo will do the rest with `topydo -f foobar.txt`.

I'll work on new tests (old ones should still be valid) and update synopsis once You'll review that enhancement.

### DIFF
--- a/topydo/commands/AddCommand.py
+++ b/topydo/commands/AddCommand.py
@@ -20,6 +20,7 @@ from datetime import date
 import re
 from sys import stdin
 import codecs
+from os.path import expanduser
 
 from topydo.lib.Config import config
 from topydo.lib.Command import Command
@@ -43,7 +44,7 @@ class AddCommand(Command):
 
         for opt, value in opts:
             if opt == '-f':
-                self.from_file = value
+                self.from_file = expanduser(value)
 
         self.args = args
 
--- a/topydo/commands/AddCommand.py
+++ b/topydo/commands/AddCommand.py
@@ -36,7 +36,6 @@ class AddCommand(Command):
         super(AddCommand, self).__init__(
             p_args, p_todolist, p_out, p_err, p_prompt)
         self.text = ' '.join(p_args)
-        self.todo = None
         self.from_file = None
 
     def _process_flags(self):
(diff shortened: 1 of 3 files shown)


topydo/commands/AddCommand.py
```diff
@@ -34,14 +34,26 @@ def __init__(self, p_args, p_todolist,
             p_args, p_todolist, p_out, p_err, p_prompt)
         self.text = ' '.join(p_args)
         self.todo = None
+        self.from_file = False
 
-    def _preprocess_input_todo(self):
+    def _process_flags(self):
+        opts, args = self.getopt('f')
```
bram85: I would have done this slightly differently: make `from_file` a `None` by default or the filename if given on the commandline with `-f`. So `getopt('f:')` and assign `self.from_file = value` below on line 44. Then you don't need to refer to `self.args[0]` on line 120, which may go wrong if the user neglects to pass on a filename. It's nicer to let `getopt` check whether the invocation is OK.

topydo/commands/AddCommand.py
```diff
@@ -83,20 +95,37 @@ def add_dependencies(p_tag):
 
         self.todo.set_creation_date(date.today())
 
+    def get_todos_from_file(self, p_filename):
```
bram85: While we're at it, we may also support stdin if `p_filename == '-'`.

bram85: So you get multiline adding basically for free:

```bash
$ topydo add -f - <<END
(C) Stuff
(D) More stuff due:1y
END
```

bram85:
Generally speaking I like the addition.

topydo/commands/AddCommand.py
```diff
@@ -83,23 +98,47 @@ def add_dependencies(p_tag):
...
+        else:
+            f = codecs.open(self.from_file, 'r', encoding='utf-8')
+
+        todos = f.read().splitlines()
+
+        return todos
+
+    def _add_todo(self, p_todo_text):
+        todo_text = self._preprocess_input_todo(p_todo_text)
+        self.todo = self.todolist.add(todo_text)
```
bram85: It looks like we can get rid of `self.todo` if you'd pass the `Todo` instance returned from `self.todolist.add()` to `self._postprocess_input_todo()` as a parameter.
Reason why I'd like to see it eliminated is that `AddCommand` is not necessarily dealing with a single todo item anymore, while the `todo` member attribute seems to imply otherwise. This makes reading the post-processing code a bit confusing.

Moreover, the preprocess and post-process methods do not need to be methods anymore. You can move them inside the `_add_todo` method, or outside the class.

mruwek: > It looks like we can get rid of self.todo if you'd pass the Todo instance returned from self.todolist.add() to self._postprocess_input_todo() as a parameter.

Agree.

> Moreover, the preprocess and post-process methods do not need to be methods anymore.

I agree about `_preprocess_input_todo`, but `_postprocess_input_todo` still uses `self.todolist`.

PR #34, opened 2015-05-27
Reload todolist after `edit` in prompt-mode

### LINKED ISSUE #33: Direct todo.txt editing with $EDITOR doesn't work in prompt-mode
```
topydo prompt
topydo> ls
topydo> edit
```

Add, remove or modify some todos in `$EDITOR`.

```
topydo> ls
```

Result: nothing has changed.

I tried adding `self.todolist.set_dirty()` before calling the editor, but it didn't work. CLI-mode (`topydo edit`) is not affected by this bug. I believe we need to do some sort of todo.txt reload in this case with prompt-mode.

mruwek:
This fixes #33

### DIFF
--- a/topydo/cli/Prompt.py
+++ b/topydo/cli/Prompt.py
@@ -69,6 +69,9 @@ class PromptApplication(CLIApplicationBase):
 
             try:
                 if self._execute(subcommand, args) != False:
+                    if self.todolist.is_obsolete():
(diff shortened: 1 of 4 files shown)


bram85:
I took a different (more generic) approach. Now any change in the background will be picked up and reloaded. No need to equip subcommands with knowledge about obsoleting a todo list.

mruwek:
Oh, this is much much better approach.

PR #36, opened 2015-05-29
Import prompt-toolkit only if it's needed.

### LINKED ISSUE #35: ImportError for prompt_toolkit when not installed.
prompt_toolkit is supposed to be an optional dependency, but it became a hard dependency with the UILoader (PR #26). This results in an `ImportError`, which should be handled more gracefully.

mruwek:
Also display more readable error message in case of absence of
prompt-toolkit.

This fixes #35


bram85:
Thank you.

Issue #37, opened 2015-05-30
Path relative to ~ not accepted


robertvanbregt:
If you provide a path to a filename starting with `~` (e.g. `~/todo.txt`) topydo will not do anything. Only a full path will work (`/User/<user>/todo.txt`)

(In stable version 0.3.2 using Mac OS X 10.9.5)

I will submit a pull request.

robertvanbregt:
If I add `os.path.expanduser` to `todotxt()` in Config.py, the path is not expanded. :(

    def todotxt(self):
        return os.path.expanduser(self.cp.get('topydo', 'filename'))

    def archive(self):
        return os.path.expanduser(self.cp.get('topydo', 'archive_filename'))

bram85:
Thanks for pointing out.

I'm puzzled why the call to `expanduser()` doesn't work from Config.py. I tried it myself (on Linux) and it works for me. That is when I put this in my configuration file:

```
filename = ~/todo.txt
```

To me the snippet you posted above seems to be the correct fix.

Pull request #38 is not complete, because `Config.todotxt()` and `Config.archive()` are called from multiple places, not just from `Main.py` where you addressed the issue (`EditCommand` also calls these functions).

So I need a bit more information why the solution above doesn't work for you.

mruwek:
> To me the snippet you posted above seems to be the correct fix.

Same here.

Could You apply this snippet again to `Config.py` and test it in interactive python shell and show the output here? For example

```
>>> from topydo.lib.Config import config
>>> a = config()
>>> a.todotxt()
```

robertvanbregt:
Todo.txt:

    $ cat ~/.todo.txt
    Buy milk 
    Polish shoes
    Do laundry

Config:

    $ cat ~/.topydo
    [topydo]
    filename = ~/todo.txt

Git:

    $ git status
    On branch stable
    Your branch is up-to-date with 'origin/stable'.
    nothing to commit, working directory clean 

Python:

    $ python
    >>> from topydo.lib.Config import config
    >>> a = config()
    >>> a.todotxt()
    '~/todo.txt'

Snippet applied:

    $ python
    >>> from topydo.lib.Config import config
    >>> a = config()
    >>> a.todotxt()
    '/Users/robert/todo.txt'

Works as expected.

When I run `python topydo/cli/Main.py` to try the snippet in the application, I get an error message:

	Traceback (most recent call last):
	  File "topydo/cli/Main.py", line 83, in <module>
[...]

PR #38, opened 2015-05-30
added os.path.expanduser to filenames

robertvanbregt:
Fix for #37.

### DIFF
--- a/topydo/lib/Config.py
+++ b/topydo/lib/Config.py
@@ -108,10 +108,10 @@ class _Config:
             return self.defaults['highlight_projects_contexts'] == '1'
 
     def todotxt(self):
-        return self.cp.get('topydo', 'filename')
+        return os.path.expanduser(self.cp.get('topydo', 'filename'))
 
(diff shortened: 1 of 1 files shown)


bram85:
As explained in issue #37, this fix is not complete. Ideally it should be fixed in the `Config` class.

robertvanbregt:
Applied patch to Config.py instead of Main.py

mruwek:
Can You squash your commits, so we don't unnecessarily pollute git history?

Try this:
```
git rebase -i b6796c1b00c9e37e5bcc31f54fddd8cda7931763
```

Then change `pick` to `squash` for: 0fde5b6 and 0fde5b6, edit final-commit message and after successful rebase do:

```
git push --force origin fix/issue37
```

robertvanbregt:
At your service. I'm a git newbie / n00b but learning rapidly. Sorry for the inconvenience.

bram85:
Thanks for the fix.

Issue #21, opened 2015-05-16
Proper Unicode support


mruwek:
With undergoing work towards porting topydo to python3 and supporting [jonathanslenders/python-prompt-toolkit](https://github.com/jonathanslenders/python-prompt-toolkit) we need to fix all errors with unicode decoding and encoding.

mruwek:
`topydo ical` returns this:
```
Traceback (most recent call last):
  File "/home/jacek/.local/bin/topydo", line 9, in <module>
    load_entry_point('topydo==0.3', 'console_scripts', 'topydo')()
  File "/home/jacek/.local/lib/python2.7/site-packages/topydo/cli/UILoader.py", line 39, in main
    CLIApplication().run()
  File "/home/jacek/.local/lib/python2.7/site-packages/topydo/cli/CLI.py", line 56, in run
    if self._execute(subcommand, args) == False:
  File "/home/jacek/.local/lib/python2.7/site-packages/topydo/cli/CLIApplicationBase.py", line 200, in _execute
    if command.execute() != False:
  File "/home/jacek/.local/lib/python2.7/site-packages/topydo/commands/IcalCommand.py", line 47, in execute
    return super(IcalCommand, self).execute()
  File "/home/jacek/.local/lib/python2.7/site-packages/topydo/commands/ListCommand.py", line 67, in execute
    self._print()
  File "/home/jacek/.local/lib/python2.7/site-packages/topydo/commands/IcalCommand.py", line 35, in _print
    self.out(str(self._view()))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0107' in position 146: ordinal not in range(128)
```

bram85:
Are you confident we can close this one? Haven't seen Unicode-bugs for a while.

mruwek:
`ls -f ical`, `ls -f json`, `edit`, `append` and `add` all work well. Those are in my opinion the most susceptible commands for possible UnicodeErrors.

 +1 from me to closing this one.

PR #39, opened 2015-05-30
Fix unicode in ical

mruwek:
Connected to #21

### DIFF
--- a/test/IcalCommandTest.py
+++ b/test/IcalCommandTest.py
@@ -21,7 +21,7 @@ import unittest
 
 from topydo.lib.Config import config
 from topydo.commands.IcalCommand import IcalCommand
-from test.CommandTest import CommandTest
+from test.CommandTest import CommandTest, utf8
 from test.TestFacilities import load_file_to_todolist
 
 IS_PYTHON_32 = (sys.version_info.major, sys.version_info.minor) == (3, 2)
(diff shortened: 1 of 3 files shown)


test/IcalCommandTest.py
```diff
@@ -80,5 +80,31 @@ def test_help_python32(self):
...
         self.assertEqual(self.errors, "icalendar is not supported in this Python version.\n")
 
+class IcalCommandUnicodeTest(CommandTest):
+    def setUp(self):
+        super(IcalCommandUnicodeTest, self).setUp()
+        self.todolist = load_file_to_todolist("test/data/ListCommandUnicodeTest.txt")
+
+    @unittest.skipIf(IS_PYTHON_32, "icalendar is not supported for Python 3.2")
+    def test_ical_unicode(self):
+        def replace_ical_tags(p_text):
```
bram85: Please move this function outside the class, and call it from here and from `IcalCommandTest.test_ical()`. This reduces code duplication.

test/IcalCommandTest.py
```diff
@@ -80,5 +80,31 @@ def test_help_python32(self):
...
+            # replace identifiers with dots, since they're random.
+            result = re.sub(r'\bical:....\b', 'ical:....', p_text)
+            result = re.sub(r'\bUID:....\b', 'UID:....', result)
+
+            return result
+
+        command = IcalCommand([""], self.todolist, self.out, self.error)
+        command.execute()
+
+#        self.assertTrue(self.todolist.is_dirty())
```
bram85: Should be `self.assertFalse(self.todolist.is_dirty())`

mruwek: Of course it should. I've commented it out to test something else, and then forgotten about it. Sorry.

mruwek: Wait, shouldn't it be rather `self.assertTrue(self.todolist.is_dirty())`? - ical adds `ical:` tag here after all.

test/IcalCommandTest.py
mruwek: Done.

test/IcalCommandTest.py
bram85: Good point.

PR #45, opened 2015-06-02
Safe 16 base colors

mruwek:
- Colors specified by name will return safe ANSI codes. Like `\033[0;31m` for normal red and `\033[0;1;31m` for light-red.
- Colors specified by number will return ANSI code from xterm 256 color chart like: `\033[0;38;5;1m` for red and `\033[0;38;5;9m` for light-red.
- Defaults are also *safe*.

### DIFF
--- a/topydo/lib/Colors.py
+++ b/topydo/lib/Colors.py
@@ -28,11 +28,13 @@ class Colors(object):
         self.metadata_color = config().metadata_color()
         self.link_color = config().link_color()
 
-    def _int_to_ansi(self, p_int, p_decorator='normal'):
+    def _int_to_ansi(self, p_int, p_decorator='normal', p_safe=True):
         """
         Returns ansi code for color based on xterm color id (0-255) and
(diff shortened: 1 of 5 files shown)


topydo/lib/Colors.py
```diff
@@ -28,7 +28,7 @@ def __init__(self):
         self.metadata_color = config().metadata_color()
         self.link_color = config().link_color()
 
-    def _int_to_ansi(self, p_int, p_decorator='normal'):
+    def _int_to_ansi(self, p_int, p_decorator='normal', p_safe=False):
```
bram85: I think the default for `p_safe` should be `True`.

Please also update the docstring to explain the notion of safe.

bram85:
Are there still test cases left for checking the 256 colors? I see you converted all test values to safe ANSI values, but I couldn't see a case for checking the conversion for color codes >=16.

mruwek:
> Are there still test cases left for checking the 256 colors?

Yes, for every `[colorscheme]` parameter at least one.

PR #47, opened 2015-06-12
Introduce backup feature and `revert` command

mruwek:
This is still a work in progress.

Right now after execution of each command backup is saved. With `topydo undo` it is possible to revert  **todo.txt** and **done.txt** to **latest** saved version.

Things TODO:

- [x] Properly handle archive file
- [x] Limit number of saved backups
- [x] Clean backup after successful execution of `undo`
- [x] Give some meaningful output
- [x] Docstrings
- [x] Synopsis
- [x] More tests

### DIFF
--- a/topydo.conf
+++ b/topydo.conf
@@ -6,6 +6,7 @@ default_command             = ls
 ; archive_filename          = done.txt
 colors                      = 1
 identifiers                 = linenumber ; or: text
+backup_count                = 5
 
 [add]
 auto_creation_date          = 1
--- a/topydo/cli/CLIApplicationBase.py
+++ b/topydo/cli/CLIApplicationBase.py
@@ -57,6 +57,7 @@ Available commands:
 * listprojects (lsprj)
 * postpone
 * pri
+* revert
 * sort
 * tag
 
@@ -200,7 +201,7 @@ class CLIApplicationBase(object):
         Execute a subcommand with arguments. p_command is a class (not an
         object).
         """
-        cmds_wo_backup = tuple(cmd + 'Command' for cmd in READ_ONLY_COMMANDS
+        cmds_wo_backup = tuple(cmd + 'Command' for cmd in ('Revert', ) + READ_ONLY_COMMANDS)
         if config().backup_count() > 0 and p_command and not p_command.__module__.endswith(cmds_wo_backup):
             call = [p_command.__module__.lower()[16:-7]] + p_args # strip "topydo.commands" and "Command"
             self.backup = ChangeSet(self.todolist, p_call=call)
--- a/topydo/cli/CLIApplicationBase.py
+++ b/topydo/cli/CLIApplicationBase.py
@@ -25,6 +25,7 @@ from six import PY2
 from six.moves import input
 
 MAIN_OPTS = "ac:d:ht:v"
+READ_ONLY_COMMANDS = ('List', 'ListContext', 'ListProject')
 
 
 def usage():
@@ -104,6 +105,7 @@ from topydo.commands.SortCommand import SortCommand
 from topydo.lib import TodoFile
 from topydo.lib import TodoList
 from topydo.lib import TodoListBase
+from topydo.lib.ChangeSet import ChangeSet
 from topydo.lib.Utils import escape_ansi
 
 
@@ -119,6 +121,7 @@ class CLIApplicationBase(object):
         self.todolist = TodoList.TodoList([])
         self.todofile = None
         self.do_archive = True
+        self.backup = None
 
     def _usage(self):
         usage()
@@ -170,6 +173,9 @@ class CLIApplicationBase(object):
         archive_file = TodoFile.TodoFile(config().archive())
         archive = TodoListBase.TodoListBase(archive_file.read())
 
+        if self.backup:
+            self.backup.add_archive(archive)
+
         if archive:
             command = ArchiveCommand(self.todolist, archive)
             command.execute()
@@ -194,6 +200,11 @@ class CLIApplicationBase(object):
         Execute a subcommand with arguments. p_command is a class (not an
         object).
         """
+        cmds_wo_backup = tuple(cmd + 'Command' for cmd in READ_ONLY_COMMANDS
+        if config().backup_count() > 0 and p_command and not p_command.__module__.endswith(cmds_wo_backup):
+            call = [p_command.__module__.lower()[16:-7]] + p_args # strip "topydo.commands" and "Command"
+            self.backup = ChangeSet(self.todolist, p_call=call)
+
         command = p_command(
             p_args,
             self.todolist,
@@ -222,7 +233,12 @@ class CLIApplicationBase(object):
             if config().keep_sorted():
                 self._execute(SortCommand, [])
 
+            if self.backup:
+                self.backup.save(self.todolist)
+
             self.todofile.write(self.todolist.print_todos())
+
+        self.backup = None
 
     def run(self):
         raise NotImplementedError
--- a/topydo/lib/ChangeSet.py
+++ b/topydo/lib/ChangeSet.py
@@ -1,0 +1,179 @@
+# Topydo - A todo.txt client written in Python.
+# Copyright (C) 2014 - 2015 Bram Schoenmakers <me@bramschoenmakers.nl>
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+""" This module serves for managing todo and archive changesets. """
+
+import json
+import time
+import zlib
+
+from copy import deepcopy
+from hashlib import sha1
+from os import path
+
+from topydo.lib.Config import config
+from topydo.lib.TodoList import TodoList
+
+def hash_todolist(p_todolist):
+    """ Calculates hash for TodoList.TodoList object. """
+    todolist_hash = sha1(p_todolist.print_todos().encode('utf-8')).hexdigest()
+
+    return todolist_hash
+
+def get_backup_path():
+    """ Returns full path and filename of backup file """
+    dirname, filename = path.split(path.splitext(config().todotxt())[0])
+    filename = '.' + filename + '.bak'
+
+    return path.join(dirname, filename)
+
+class ChangeSet(object):
+    """ Class for operations related with backup management. """
+
+    def __init__(self, p_todolist=None, p_archive=None, p_call=[]):
+        self.todolist = deepcopy(p_todolist)
+        self.archive = deepcopy(p_archive)
+        self.timestamp = str(int(time.time()))
+        self.call = ' '.join(p_call)
+
+        try:
+            self.json_file = open(get_backup_path(), 'r+b')
+        except IOError:
+            self.json_file = open(get_backup_path(), 'w+b')
+
+        self._read()
+
+    def _read(self):
+        """
+        Reads backup file from json_file property and sets backup_dict property
+        with data decompressed and deserialized from that file. If no usable
+        data is found backup_dict is set to the empty dict.
+        """
+        self.json_file.seek(0)
+        try:
+            data = zlib.decompress(self.json_file.read())
+            self.backup_dict = json.loads(data.decode('utf-8'))
+        except (EOFError, zlib.error):
+            self.backup_dict = {}
+
+    def _write(self):
+        """
+        Writes data from backup_dict property in serialized and compressed form
+        to backup file pointed in json_file property.
+        """
+        self.json_file.seek(0)
+        self.json_file.truncate()
+        dump = json.dumps(self.backup_dict)
+        dump_c = zlib.compress(dump.encode('utf-8'))
+        self.json_file.write(dump_c)
+
+    def add_archive(self, p_archive):
+        """ Sets deep copy of p_archive as archive attribute. """
+        self.archive = deepcopy(p_archive)
+
+    def add_todolist(self, p_todolist):
+        """ Sets deep copy of p_todolist as todolist attribute. """
+        self.todolist = deepcopy(p_todolist)
+
+    def save(self, p_todolist):
+        """
+        Saves a tuple with archive, todolist and command with its arguments
+        into the backup file with unix timestamp as the key. Tuple is then
+        indexed in backup file with combination of hash calculated from
+        p_todolist and unix timestamp. Backup file is closed afterwards.
+        """
+        self._trim()
+
+        current_hash = hash_todolist(p_todolist)
+        list_todo = (self.todolist.print_todos()+'\n').splitlines(True)
+        list_archive = (self.archive.print_todos()+'\n').splitlines(True)
+
+        self.backup_dict[self.timestamp] = (list_todo, list_archive,  self.call)
+
+        index = self._get_index()
+        index.insert(0, (self.timestamp, current_hash))
+        self._save_index(index)
+
+        self._write()
+        self.close()
+
+    def delete(self, p_timestamp=None):
+        """ Removes backup from the backup file. """
+        timestamp = p_timestamp or self.timestamp
+        index = self._get_index()
+
+        try:
+            del self.backup_dict[timestamp]
+            index.remove(index[[change[0] for change in index].index(timestamp)])
+            self._save_index(index)
+            self._write()
+        except KeyError:
+            pass
+
+    def _get_index(self):
+        try:
+            index = self.backup_dict['index']
+        except KeyError:
+            self.backup_dict['index'] = []
+            index = self.backup_dict['index']
+
+        return index
+
+    def _save_index(self, p_index):
+        """
+        Saves index of backups supplied in p_index into the backup_file
+        property with 'index' as the key.
+        """
+        self.backup_dict['index'] = p_index
+
+    def _trim(self):
+        """
+        Removes oldest backups that exceed the limit configured in backup_count
+        option.
+        """
+        index = self._get_index()
+        backup_limit = config().backup_count() - 1
+
+        for changeset in index[backup_limit:]:
+            self.delete(changeset[0])
+
+    def get_backup(self, p_todolist):
+        """
+        Retrieves a backup for p_todolist from backup file and sets todolist,
+        archive and call attributes to appropriate data from it.
+        """
+        change_hash = hash_todolist(p_todolist)
+
+        index = self._get_index()
+        self.timestamp = index[[change[1] for change in index].index(change_hash)][0]
+
+        d = self.backup_dict[self.timestamp]
+
+        self.todolist = TodoList(d[0])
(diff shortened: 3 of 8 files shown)


topydo.conf
```diff
@@ -6,6 +6,7 @@ default_command             = ls
 ; archive_filename          = done.txt
 colors                      = 1
 identifiers                 = linenumber ; or: text
+; backup                    = 1
```
bram85: I would combine the two new options. If `backup` is set, then make backups. If `0` or empty, don't.

topydo/cli/CLIApplicationBase.py
```diff
@@ -101,6 +101,8 @@ def version():
 from topydo.lib import TodoList
 from topydo.lib import TodoListBase
 from topydo.lib.Utils import escape_ansi
+from topydo.lib.Colors import Colors
```
bram85: This import is not needed.

topydo/lib/Config.py
```diff
@@ -126,6 +130,12 @@ def archive(self):
     def identifiers(self):
         return self.cp.get('topydo', 'identifiers')
 
+    def backup(self):
+        try:
+            return self.cp.getboolean('topydo', 'backup')
+        except ValueError:
+            return self.defaults['colors'] == '1'
```
bram85: Wrong key, should be `backup`.

topydo/commands/UndoCommand.py
```diff
@@ -0,0 +1,54 @@
...
+            return False
+
+        archive_file = TodoFile.TodoFile(config().archive())
+        archive = TodoList.TodoList(archive_file.read())
+
+        last_change = ChangeSet()
+        try:
+            last_change.read(self.todolist)
+        except KeyError:
+            self.error('No backup were found for current state of ' + config().todotxt())
```
bram85: *"No backup was found for the current state of ..."*

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,94 @@
...
+        current_hash = self._hash_todolist(p_todolist)
+        self.shelve_file[current_hash] = (self.todolist, self.todolist_hash, self.archive, self.archive_hash)
+
+        self.close()
+
+    def delete(self, p_todolist_hash):
+        """ Removes backup indexed with p_todolist_hash from backup file. """
+        try:
+            del self.shelve_file[p_todolist_hash]
+        except:
```
bram85: `except KeyError`

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if self.archive:
+            p_archive.erase()
+            p_archive.add_todos(self.archive.todos())
+
+    def _hash_todolist(self, p_todolist):
+        """ Calculates hash for TodoList.TodoList object. """
+        try:
+            todolist_hash = sha1(p_todolist.print_todos().encode('utf-8')).hexdigest()
+        except:
```
bram85: What kind of exception(s) are you trying to catch here?

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,94 @@
...
+        Saves a tuple with archive, todolist and their hashes into the
+        backup file. Tuple is indexed in backup file  with hash calculated from
+        p_todolist.
+        """
+        current_hash = self._hash_todolist(p_todolist)
+        self.shelve_file[current_hash] = (self.todolist, self.todolist_hash, self.archive, self.archive_hash)
+
+        self.close()
+
+    def delete(self, p_todolist_hash):
```
bram85: It's not consistent that all other methods in this class accept a `TodoList` instance, while this method takes a hash. I would make this accepting a `TodoList` as well.

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,94 @@
...
+
+        self.todolist = d[0]
+        self.todolist_hash = d[1]
+        self.archive = d[2]
+        self.archive_hash = d[3]
+
+    def apply(self, p_todolist, p_archive):
+        """ Applies changeset on supplied p_todolist. """
+        if self.todolist:
+            p_todolist.erase()
```
bram85: I think a `TodoListBase.replace(self, p_todos)` would be nice to erase it call `add_todos`. `SortCommand` also uses this pattern (see also commit 4147d78).

topydo.conf
bram85: Or on second thought, make it just a boolean and eliminate the `backup_file` option. And make the location always relative to the `todo.txt` file (same directory, but something like `.filename.bak`, where filename is the value of the `filename` option.

The motivation is to prevent that a changeset contains revisions of other todo.txt files than the currently configured one. Having a different revision on top of the undo stack would break the `undo` command.

bram85:
* Would a `redo` be feasible?
* Store along with the file also the command (as a string) that is being undone, to improve user feedback. You could use that for a (future) `undo ls` - or something - to show a history of commands.
* As a test suggestion, which you may already have in mind, is a case where the `todo.txt` file is modified between some command and `undo`.

kovetskiy:
Why backup directory, why not git repository? + With git `redo` will be feasible.

bram85:
@kovetskiy A valid question, but I'd like to counter this proposal for various reasons.

I'd like to keep topydo light of hard dependencies, especially on external commands. It sounds to me a lot of overhead to create a git commit with every change I make to the todo.txt file (and the done.txt file for that matter).

Also, a git repository has an infinite history, no way to purge the older versions that are no longer relevant. Although it will be cheap in disk-space with the way git stores similar objects.

And what if the todo.txt and done.txt file are already in a Git repo, but part of some project? (as already is the case in a single instance I've seen on Github). A submodule for the todo.txt and done.txt?

kovetskiy:
@bram85 worries about disk-space in 2015 it is not seriously. With git you can purge old commits, just using `rebase` with `fixup` old commits, and running `git gc`.

And, of course, todo.txt and done.txt can be in some repository, topydo can create a directory with git repository (like a backup directory) and sync files with this repository.

I agree about overhead, but this gives opportunity to use `undo` or `redo` more than one time.

bram85:
I'm still not a fan of baking in git into topydo. It's up to @mruwek what he thinks is best, after all, he's the one who implements the undo functionality. But I'm fine with the current approach he's taking, it's simple and does the job.

As an alternative you could write a wrapper bash script around topydo that commits, undoes and redoes. topydo is a command line tool after all. I'd be willing to add something like that in a

topydo/cli/CLIApplicationBase.py
mruwek: Done in 7ba7ae30a77f65d81546eb1c3c0feccb6cb2249b

topydo/lib/Config.py
mruwek: Done in 7dd7c3c90926f6f2e6c2deab2093b657cf434552

topydo/commands/UndoCommand.py
mruwek: Fixed in 3f88d28c7449bcd32c46f52010d6360decaaae7f

topydo/lib/ChangeSet.py
mruwek: Done in d3d3563ce1d9036fae91c431aa880cb11ac6b23f

topydo/lib/ChangeSet.py
mruwek: AttributeError. Done in b44296ee8bf3adc50facd3eed496c45a66ba2580

topydo/lib/ChangeSet.py
mruwek: Done in: d3d3563ce1d9036fae91c431aa880cb11ac6b23f

topydo/lib/ChangeSet.py
mruwek: Done in: d3d3563ce1d9036fae91c431aa880cb11ac6b23f

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,94 @@
...
+import shelve
+from copy import deepcopy
+
+class ChangeSet(object):
+    def __init__(self, p_todolist=None, p_archive=None):
+        self.todolist = deepcopy(p_todolist)
+        self.todolist_hash = self._hash_todolist(p_todolist)
+        self.archive = deepcopy(p_archive)
+        self.archive_hash = self._hash_todolist(p_archive)
+        self.shelve_file = shelve.open(config().backup_file())
```
bram85: The path of the backup file should be relative to the active todo.txt file. Otherwise things get messy with multiple todo.txt files, or when running topydo from different working directories.

topydo/commands/UndoCommand.py
```diff
@@ -0,0 +1,56 @@
...
+
+        last_change.close()
+
+
+    def usage(self):
+        return """Synopsis: undo"""
+
+    def help(self):
+        return """\
+Undoes last command.
```
bram85: _Undoes the last command_

test/UndoCommandTest.py
```diff
@@ -0,0 +1,125 @@
...
+        todos = [
+            "Foo",
+            "Bar",
+            "Baz",
+        ]
+
+        self.todolist = TodoList(todos)
+        self.today = date.today()
+
+        self.tmp_name = str(uuid4().hex.upper()[0:6])
```
bram85: Python has a _tempfile_ module for stuff like this.

test/UndoCommandTest.py
```diff
@@ -0,0 +1,125 @@
...
+
+        archive_filename = '/tmp/' + self.tmp_name + '_archive'
+        backup_filename = '/tmp/' + self.tmp_name + '_backup'
+
+        config(p_overrides={('backup', 'backup_file'): backup_filename})
+        config(p_overrides={('topydo', 'archive_filename'): archive_filename})
+
+        self.archive_file = TodoFile(archive_filename)
+        #self.archive = TodoList(self.archive_file.read())
```
bram85: Remove this line.

bram85:
@mruwek Can you indicate which TODO items in the top post are addressed now? As far as I can see, limiting the number of stored backups is not implemented yet, right?

topydo.conf
mruwek: done in ff953b8fab8c3523efc4799721e824c669039a1a

topydo/lib/ChangeSet.py
mruwek: done in ff953b8fab8c3523efc4799721e824c669039a1a

topydo/commands/UndoCommand.py
mruwek: fixed in dde712d74507d50f12967ce4c57bd2a24e06473c

test/UndoCommandTest.py
mruwek: I know, but this way it seems easier to remove those files in `tearDown()`

test/UndoCommandTest.py
mruwek: done in 992608af03f90dd8bf6c2a988c32c7a56d19b87f

topydo/lib/ChangeSet.py
mruwek: I had to revert it in ad2fe7443c56a0c1caa53c632109c54d2a18a101 because this way it is much more simple to remove Changeset after it has been used in `undo` (feccf937b39a59428c80c493d760a87db6602e51). Otherwise we would have to make another `deepcopy` of `TodoList`.

mruwek:
I perceive this as feature-complete right now. I'll write docstrings, synopsis and think about output after your review.

bram85:
Thanks. Give me some days to have a look at it (it's a bit much to look at on a Monday morning just before work :) )

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,133 @@
...
+        self.close()
+
+    def delete(self, p_todolist_hash):
+        """ Removes backup indexed with p_todolist_hash from backup file. """
+        index = self.get_index()
+
+        try:
+            del self.shelve_file[p_todolist_hash]
+            index.remove(p_todolist_hash)
+        except KeyError:
```
bram85: Coverage reveals that there's no test that covers this exception.

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def apply(self, p_todolist, p_archive):
+        """ Applies changeset on supplied p_todolist. """
+        if self.todolist:
+            p_todolist.replace(self.todolist.todos())
+
+        if self.archive:
+            p_archive.replace(self.archive.todos())
+
+    def _hash_todolist(self, p_todolist):
```
bram85: Could be a function outside `ChangeSet` since `self` is not used.

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,133 @@
...
+        """ Calculates hash for TodoList.TodoList object. """
+        try:
+            todolist_hash = sha1(p_todolist.print_todos().encode('utf-8')).hexdigest()
+        except AttributeError:
```
bram85: To me it's not apparent what could trigger the `AttributeError`.

topydo/cli/CLIApplicationBase.py
```diff
@@ -25,6 +25,7 @@
 from six.moves import input
 
 MAIN_OPTS = "ac:d:ht:v"
+NON_DESTRUCTIVE_COMMANDS = ('List', 'ListContext', 'ListProject')
```
bram85: Suggestion is to rename to `READ_ONLY_COMMANDS`. I don't consider all other commands to be destructive, they just do what they're supposed to do ;)

bram85:
I guess it's looking quite good (didn't have time yet to check the test cases). I'll give it a spin this week and see if I encounter any issues with it.

topydo/cli/CLIApplicationBase.py
```diff
@@ -189,6 +195,10 @@ def _execute(self, p_command, p_args):
         Execute a subcommand with arguments. p_command is a class (not an
         object).
         """
+        cmds_wo_backup = tuple(cmd + 'Command' for cmd in ('Undo', ) + NON_DESTRUCTIVE_COMMANDS)
+        if config().backup() and not p_command.__module__.endswith(cmds_wo_backup):
```
bram85: Got a crash on this line when running the `help` subcommand in prompt mode:

```
topydo> help
  File "/home/brschoen/.local/bin/topydo", line 11, in <module>
    sys.exit(main())
  File "/home/brschoen/.local/lib/python3.4/site-packages/topydo/cli/UILoader.py", line 42, in main
    PromptApplication().run()
  File "/home/brschoen/.local/lib/python3.4/site-packages/topydo/cli/Prompt.py", line 110, in run
    if self._execute(subcommand, args) != False:
  File "/home/brschoen/.local/lib/python3.4/site-packages/topydo/cli/CLIApplicationBase.py", line 199, in _execute
    if config().backup() and not p_command.__module__.endswith(cmds_wo_backup):
AttributeError: 'NoneType' object has no attribute '__module__'
```

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,133 @@
...
+
+        try:
+            del self.shelve_file[p_todolist_hash]
+            index.remove(p_todolist_hash)
+        except KeyError:
+            pass
+
+        self.save_index(index)
+
+    def get_index(self):
```
bram85: Shouldn't this be an internal function? I.e.: `_get_index`. Same for `save_index` below.

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,133 @@
...
+
+        self.close()
+
+    def delete(self, p_todolist_hash):
+        """ Removes backup indexed with p_todolist_hash from backup file. """
+        index = self.get_index()
+
+        try:
+            del self.shelve_file[p_todolist_hash]
+            index.remove(p_todolist_hash)
```
bram85: I see duplicate entries in the `shelve_file['index']`, as a result of adding a todo and immediately removing it with `del`. The resulting todo file is back to the original state, so it hashes to the same value again. So this `index.remove` is not guaranteed to remove only one value from the index.

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,133 @@
...
+    def save(self, p_todolist):
+        """
+        Saves a tuple with archive, todolist and their hashes into the
+        backup file. Tuple is indexed in backup file  with hash calculated from
+        p_todolist.
+        """
+        self.trim()
+
+        current_hash = self._hash_todolist(p_todolist)
+        self.shelve_file[current_hash] = (self.todolist, self.todolist_hash, self.archive, self.archive_hash)
```
bram85: Is it still necessary to store `self.todolist_hash` and `self.archive_hash`? Unless I'm overlooking something, these variables don't seem to be used (anymore).

bram85:
After having backups enabled for one morning, I notice that the file size of the backup file keeps increasing quite a bit. Actually, with each operation, 512 KB is added to the file (I have a fairly large todo.txt and done.txt file, combined).

It's most likely due to the backend chosen by the `shelve` module. According to [this post](https://mail.python.org/pipermail/python-list/2004-November/260572.html), it acts as a database, which allocates space as it pleases. When deleting a key, the data is not actually removed. I verified this, the data is still there even after removing its key. Only when a key is reused it, that space is reused. Which is unlikely in our case since the keys are hashes of the todo list.

These size increases are a show-stopper in my opinion, 512 KB for each operation is a bit excessive. My suggestions:

* Use a static set of keys, e.g. 0, 1, 2, 3 and 4 and shift their values with each operation (and discard the value in slot 4).
* Perhaps shelve could use a different backend, but this may affect portability. Not sure if this solves the problem.
* A different persistence module altogether? Perhaps use a plain `pickle` to dump our own dict object to a file? (I think we discussed this briefly on Gitter some months ago).

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,158 @@
...
+        timestamp = p_timestamp or self.timestamp
+        index = self._get_index()
+
+        try:
+            del self.backup_dict[timestamp]
+            index.remove(index[[change[0] for change in index].index(timestamp)])
+        except KeyError:
+            pass
+
+        self._save_index(index)
```
bram85: I think `_save_index` and `_write` should be called from within the `try:` block. With an invalid key it makes no sense to write back the data.

mruwek: done in 15da2ff93915f76026a73ab3cb9adde703e94d7d

mruwek:
History is now squashed to acceptable size. Docstrings are added/updated. Waiting for final review.

topydo.conf
```diff
@@ -6,6 +6,7 @@ default_command             = ls
 ; archive_filename          = done.txt
 colors                      = 1
 identifiers                 = linenumber ; or: text
+; backup_count              = 5
```
bram85: I think it should be uncommented. Like this, the default value (5) will silently kick in and make backups nevertheless, so it's better to be explicit about it.

topydo/lib/ChangeSet.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def _save_index(self, p_index):
+        """
+        Saves index of backups supplied in p_index into the backup_file
+        property with 'index' as the key.
+        """
+        self.backup_dict['index'] = p_index
+
+    def _trim(self):
+        index = self._get_index()
```
bram85: A docstring is missing here.

topydo.conf
mruwek: done

topydo/lib/ChangeSet.py
mruwek: I've overlooked that one. Thanks.

bram85:
Thank you!

Issue #32, opened 2015-05-26
Default coloring for list does not work on Windows 7


robertvanbregt:
When I run a `list` action, the color coding does not work. Instead of colors, I can see the equivalent characters, for example

<pre>
←[0m| 30| 2015-05-19 Buy milk ←[31m+Groceries←[0m←[0m←[0m←[31m@errands←[0m ←[31m
</pre>

robertvanbregt:
A solution would be to add a flag like `-p` or `--plain` to remove any formatting in the output?

bram85:
As a work-around, you can disable colors with a [configuration](https://github.com/bram85/topydo/wiki/Configuration) option:

```
[topydo]
colors = 0
```

mruwek:
@robertvanbregt 

You can also try to install [ansicon](https://github.com/adoxa/ansicon).

@bram85
I've investigated [colorama](https://github.com/tartley/colorama) and [termcolor](https://pypi.python.org/pypi/termcolor), but those would implement only basic 16 colors.

robertvanbregt:
The colors switch will do.

bram85:
@mruwek: also I would like to keep topydo light on mandatory requirements, so I'm not so much in favor at this moment to create such a dependency just for the colors.

Still, colors should work for Windows.

mruwek:
> Still, colors should work for Windows.

It's either colorama then or our own lib for translating ANSI codes to windoze `SetConsoleTextAttribute` calls.

mruwek:
Or this: https://code.google.com/p/colorconsole/source/browse/colorconsole/win.py

bram85:
@robertvanbregt Which console do you use where this color sequences appear in the output? Is it still an issue in the `master` branch?

robertvanbregt:
I am using topydo 0.3.2 on my fairly locked down work PC. I run Win 7 SP
1 and launch a console from the Start > Run > Cmd prompt.

I have no git here, so it requires some effort to download and run the
master branch version. When I have time, I will give it a shot.

--
Groet, Regards,
Robert van Bregt
 
 
Op Wo jun 3 2015, om 07:59 schreef Bram Schoenmakers:
> @robertvanbregt[1] Which console do you use where this color sequences appear in the output? Is it still an issue in the master branch?


> —
> Reply to this email directly or view it on GitHub[2]. 

>  
 


Links:

  1. https://github.com/robertvanbregt
[...]

PR #49, opened 2015-08-10
Windows cmd 2 -- add color to output on Windows

MinchinWeb:
I offer this as a solution to #32. [colorama](https://pypi.python.org/pypi/colorama) is used. What `colorama` does is (on Windows) is take ANSI color sequences and convert them to Win32 API calls. On other platforms, it does nothing. In this case, I have also made it so `colorama` is only installed and called in Windows systems. In my experience, `colorama` is the simplest and easiest way to add cross-platform color command-line output.

This pull request builds on #48, although no changes in that pull request should be required to make this work.

Now pictures:

## Before
before

## After
after

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,24 @@
 *.pyc
 *.sw?
-build
-dist
 install
 .coverage
+
+# Distribution / packaging
+.Python
+env/
+env*/
+build/
+develop-eggs/
+dist/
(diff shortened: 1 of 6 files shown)


bram85:
Looks great, thanks a lot for your contribution!

MinchinWeb:
Glad to help!

PR #52, opened 2015-09-27
Top command

MinchinWeb:
First off, the framework you have built up here is great. Adding this changes proved surprisingly easy. Thanks for all your work on `topydo`!.

---

This pull request puts together a series of changes to `ls` output. It:

 - unbracketed priorities
 - moves added date (and due date) to end of the line
 - reports dates in a 'human readable' way
 - indents the second lines of items, and
 - updated tests to match

Sample output:
```
C:\>topydo ls +testing
|358|   this has a due date, and was added two month ago or so, and has no
          priority +testing (2 months ago, overdue by 11 days)
|356| A test item +testing (today)
|357| D this is a really long item that goes over two lines. Yup, it keeps
          going and going and going some more +testing (today)
|359|   this item is a year old +testing (13 months ago)
|360|   this item is over two years old +testing (2 years ago)
```

## Top Command

The above was really just the prelude to introduce the `top` command. I added this because, personally, this is my most frequently used command to manage my todo list. The `top` command is like the `ls` command, but truncates output to one line per item, and one "screenful" of lines. It does that by using the new settings `console_height` and `consule_width` (which default to 25 and 80 respectively). Because it is built off of `ls`, and so can use the same filtering rules.

Sample output:
```
C:\>topydo top +testing
|358|   this has a due date, and was added two month ago or so, and has no ...
|356| A test item +testing (today)
|357| D this is a really long item that goes over two lines. Yup, it keeps ...
|359|   this item is a year old +testing (13 months ago)
|360|   this item is over two years old +testing (2 years ago)
```

Issues I've run into:
 - adding line breaks can mean that really long URL's (over ~60 chars) can end up over two lines and only the first part will be coloured
 - line breaking, wrapping, and indenting is done with the `textwrap` module, which is part of the standard library. However, it makes uses of some functionality that was introduced in Python 3.4, so a work-around for earlier versions had to be created, but seems to work fine.

I also included various tyopgraphical fixes to be more inline with PEP 8, and added Python 3.5 (which is now out) to Travis-CI.


mruwek:
+1 for human-readable-format in `topydo ls`, but I'd like this with some kind of option (`-H` seems to be quite natural), rather than by default.

MinchinWeb:
As suggested, I've made the 'human readable' dates an option for `topydo ls`, but kept them on for `topydo top` (where brevity is more of an issue). I used `-r` (for 'relative` dates) rather than `-H` as the Windows CMD is generally not case sensitive, and so I'd rather not have both `-h` and `-H` options (but I'm open to other suggestions).

I also added tests for the `top` command, based on the `ls` tests.

bram85:
Thanks a lot for your code. It's a big patch, I'll need some time to look at it.

topydo.conf
```diff
@@ -6,6 +6,8 @@ default_command             = ls
 ; archive_filename          = done.txt
 colors                      = 1
 identifiers                 = linenumber ; or: text
+console_height              = 25
```
bram85: These options are static and not really flexible, I'd suggest to read the console size dynamically at runtime. For example:

```python
rows, columns = subprocess.check_output(['stty', 'size']).split()
```

[Source](https://stackoverflow.com/questions/566746/how-to-get-console-window-width-in-python)

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -84,15 +86,40 @@ def filter(self, p_todo_str, p_todo):
...
-        return ' ' * self.indent + p_todo_str
+        try:
+            return(textwrap.fill(p_todo_str,
+                                 initial_indent=' '*self.indent,
+                                 subsequent_indent=' '*(10 + self.indent),
+                                 width=config().console_width() - 1,
+                                 break_long_words=True,  # will break long URL's
+                                 max_lines=self.max_lines,  # requires Python 3.4
+                                 placeholder=' ...'))  # requires Python 3.4
+        except:
```
bram85: Please write out which exception you're catching here.

topydo/Commands.py
```diff
@@ -32,7 +32,7 @@
     'depri': 'DepriCommand',
     'do': 'DoCommand',
     'edit': 'EditCommand',
-    'exit': 'ExitCommand', # used for the prompt
+    'exit': 'ExitCommand',  # used for the prompt
```
bram85: Please revert the change in this line. I'd like to keep history as clean as possible.

topydo/commands/ListCommand.py
```diff
@@ -124,12 +133,13 @@ def help(self):
...
 -f : Specify the output format, being 'text' (default), 'ical' or 'json'.
 
-     * 'text' - Text output with colors and identation if applicable.
+     * 'text' - Text output with colors and indentation if applicable.
      * 'ical' - iCalendar (RFC 2445). Is not supported in Python 3.2. Be aware
                 that this is not a read-only operation, todo items may obtain
                 an 'ical' tag with a unique ID. Completed todo items may be
                 archived.
      * 'json' - Javascript Object Notation (JSON)
+-r : Displays dates in a relative, human-readable format
```
bram85: Suggestion to only mention due dates and creation dates.

```
-r : Displays creation and due dates in a relative, human-readable format.
```

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -117,3 +145,119 @@ def filter(self, p_todo_str, _):
...
+        alignment)
+    """
+
+    def __init__(self, p_replacement=" "):
+        super(PrettyPrinterBasicPriorityFilter, self).__init__()
+        self.p_replacement = p_replacement
+
+    def filter(self, p_todo_str, _):
+        """ Find the priority """
+        matches = re.search('^(?P<id>\| *\w+\| )?(\((?P<pri>[A-Z])\) )?', p_todo_str)
```
bram85: Please add a comment that gives an example of a string you're trying to match. That makes parsing this regex a bit easier.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -117,3 +145,119 @@ def filter(self, p_todo_str, _):
...
+                    """ If we have a priority """
+                    return(matches.group('pri') + ' ' + p_todo_str[matches.span()[1]:])
+                else:
+                    return(self.p_replacement + ' ' + p_todo_str[matches.span()[1]:])
+        else:
+            """ No match found """
+            return(self.p_replacement + ' ' + p_todo_str)
+
+
+class PrettyPrinterHumanDatesFilter(PrettyPrinterFilter):
```
bram85: I'd prefer to see an external library such as [humanize](https://github.com/jmoiron/humanize) instead of having our own code. What do you think? Will it suit our needs?

On one hand I'd like to keep topydo light on dependencies, but we could make it an optional dependency, like prompt_toolkit and icalendar.

topydo/lib/Version.py
```diff
@@ -2,7 +2,7 @@
 
 VERSION = '0.6'
 
-LICENSE = """Copyright (C) 2014 Bram Schoenmakers
+LICENSE = """Copyright (C) 2014-15 Bram Schoenmakers
```
bram85: 2014-2015

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -117,3 +145,119 @@ def filter(self, p_todo_str, _):
                 p_todo_str)
 
         return p_todo_str
+
+
+class PrettyPrinterBasicPriorityFilter(PrettyPrinterFilter):
+
+    """ Adds strips the bracked from the priority of the todo item. Add a
```
bram85: _Strips the brackets from the priority..._

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -84,15 +86,40 @@ def filter(self, p_todo_str, p_todo):
...
         self.indent = p_indent
+        self.max_lines = p_max_lines
 
     def filter(self, p_todo_str, _):
         """ Applies the indentation. """
-        return ' ' * self.indent + p_todo_str
+        try:
+            return(textwrap.fill(p_todo_str,
+                                 initial_indent=' '*self.indent,
+                                 subsequent_indent=' '*(10 + self.indent),
```
bram85: The 10 seems to be a magic number to me, which needs a clarification in a comment.

bram85:
Left some feedback on the code, I haven't looked at the test code yet. Will do so in the next round(s) when the comments are addressed.

bram85:
I think it should also be an option in the configuration file to always enable human readable dates, rather than only a `-r` flag.

```ini
[ls]
print_relative_dates = 1
```

topydo.conf
MinchinWeb: I'm not opposed to determining console size at run time, but we need to consider:
1. what happens when we run this on Windows? `stty` isn't available, by default, on Windows systems
2. what happens if we run somewhere that isn't a terminal, like even the test suite?

I'll keep looking to see if I can find a multiplatform solution.

topydo/Commands.py
MinchinWeb: It was intentional. This fixes PEP8 E261 ("at least two spaces before inline comment").

topydo/lib/PrettyPrinterFilter.py
MinchinWeb: I'm open to using humanize ([arrow](http://crsmithdev.com/arrow/) seems like another option), but both only advertise support for Python 2.7 and 3.3. I don't like the idea of re-writing this to work with an external library, but still having to keep this code. So can we drop support for Python 3.2?

topydo/commands/ListCommand.py
MinchinWeb: Fixed in 975154064912fd999a6f46f1dff7ebd05d1b213a

topydo/lib/PrettyPrinterFilter.py
MinchinWeb: Comment added in 975154064912fd999a6f46f1dff7ebd05d1b213a

topydo/lib/PrettyPrinterFilter.py
MinchinWeb: Fixed in 975154064912fd999a6f46f1dff7ebd05d1b213a

topydo/lib/PrettyPrinterFilter.py
MinchinWeb: Oops. Fixed in 975154064912fd999a6f46f1dff7ebd05d1b213a

topydo/lib/PrettyPrinterFilter.py
MinchinWeb: Comment added in 975154064912fd999a6f46f1dff7ebd05d1b213a

topydo/lib/Version.py
MinchinWeb: Fixed in 975154064912fd999a6f46f1dff7ebd05d1b213a

topydo.conf
MinchinWeb: Is this a good solution? --> <https://gist.github.com/jtriley/1108174>

MinchinWeb:
@bram85 , I've addressed most of your comments. There are two outstanding items:

1. I can switch the date code to `humanize`, but then we should probably drop Python 3.2 support.
2. How to determine console size directly

I await your comments.

topydo.conf
MinchinWeb: [shutil.get_terminal_size()](https://docs.python.org/3/library/shutil.html#querying-the-size-of-the-output-terminal) seems like the best option, as it is part of the standard library, but is only available for Python 3.3+ (and not available for Python 2.7).

I think this is my preferred solution, but I'm not sure how best to deal with Python 2.7 and 3.2

bram85:
@MinchinWeb 

1. I tend to say to drop Python 3.2 support. The (optional) dependency `icalendar` also doesn't support it, and I made some ugly 3.2 vs. non-3.2 cases at some places. I'm not sure how widely used Python 3.2 is these days, Debian Wheezy has this version. On the other hand, people can always use Python 2.7 if they cannot upgrade their Python 3 version.
2. Needs some research how to determine this nicely, cannot answer/investigate right now. Alternative is to determine it run-time, and fallback to configured values if this fails (e.g. on Windows).

bram85:
Regarding Python 3.2, I dropped its support in commit a0a50ba951814f3f5a2080cc62e5a0a57916c6fb.

MinchinWeb:
To determine the console size, I'm using `shutil.get_terminal_size()`. For Python versions before 3.3 (so really just Python 2.7), I use a [backported version of it](https://pypi.python.org/pypi/backports.shutil_get_terminal_size). However, the tests pass on my computer (Python 2.7, Windows 10), but don't on Travis-CI. The code in question is (pulled from [here](https://github.com/MinchinWeb/topydo/blob/f1e7e5afa40ba3dc805b4d9e8a9fbec2d685801b/topydo/lib/Utils.py#l61-L64)):
```python
    try:
        from shutil import get_terminal_size
    except ImportError:
        from backports.shutil_get_terminal_size import get_terminal_size
```
On my machine, the second line throws a `ImportError`. Can someone confirm if it throwing a different error on Linux?

MinchinWeb:
Found the issue was actually with pylint rather than the tests. Fixed!

MinchinWeb:
I ended up using `arrow` for humanizing date, because it seems to be more popular that `humanize`, and `arrow` remains under active development.

@bram85 : I think I've addressed all your comments.

setup.py
```diff
@@ -36,6 +38,7 @@ def find_version(*file_paths):
     url = "https://github.com/bram85/topydo",
     install_requires = [
         'six >= 1.9.0',
+        'arrow >= 0.6.0',
```
bram85: I see that arrow 0.7.0 was released a couple of hours ago. Perhaps bump the version and verify that all works OK? I expect no problems, but it's better to qualify topydo with the latest version that users may install when installing/upgrading topydo.

test/AddCommandTest.py
```diff
@@ -200,7 +200,7 @@ def test_add_dep9(self):
         command = ListCommand.ListCommand(["Bar"], self.todolist, self.out, self.error)
         command.execute()
 
-        self.assertEqual(self.output, "|kbn| {today} Bar p:1 +Project\n|kbn| {today} Bar +Project\n".format(today=self.today))
+        self.assertEqual(self.output, "|kbn| {today} Bar p:1 +Project\n|kbn|   {today} Bar +Project\n".format(today=self.today))
```
bram85: Why these additional spaces? They are added in [this commit](https://github.com/MinchinWeb/topydo/commit/b1b1d4f8969c8d1f93399b9e5a8e52374646f28a) under "random style improvements". Same remark for the diff below.

bram85: I suspect it's related to [this commit](https://github.com/MinchinWeb/topydo/commit/4bc70e336aaeb14ced1ce7df3c5f10421da3e537)? If so, it would be best to logically combine commits: if you break a test case with a code change, add the fixed test in the same commit.

topydo.conf
```diff
@@ -14,6 +14,7 @@ auto_creation_date          = 1
 hide_tags                   = id,p,ical
 indent                      = 0
 list_limit                  = -1
+human_readable_dates		= 1
```
bram85: Minor: misaligned.

topydo/commands/TopCommand.py
```diff
@@ -0,0 +1,105 @@
...
+from topydo.lib.PrettyPrinterFilter import (
+    PrettyPrinterIndentFilter,
+    PrettyPrinterHideTagFilter,
+    PrettyPrinterBasicPriorityFilter,
+    PrettyPrinterHumanDatesFilter
+)
+from topydo.lib.Utils import get_terminal_size
+
+
+class TopCommand(ExpressionCommand):
```
bram85: This class has a lot of common with the `ListCommand`. Did you consider to subclass it from `ListCommand` and just override the `_print()` method?

The test cases are also almost identical to the `ListCommand`, so I feel there's some duplication going on here.

bram85:
* I think `arrow` is a good choice, we could use it in other places as well (I'm thinking of the [RelativeDate module](https://github.com/bram85/topydo/blob/master/topydo/lib/RelativeDate.py)). But that can be done at a later stage by anyone.

  Please document the new dependencies also in the [README.md file](https://github.com/bram85/topydo/blob/master/README.md).

* Overall, I think the new formatting of `ls` is an improvement.

* `top` also looks nice, although I'm not sure whether it should be a separate command or a flag for `ls` (e.g. `ls -1`).

* Suggestion for `top` is to always have the creation + due date visible, and truncate the todo text earlier to make space for it. With long items, I don't see whether they have a due date, or which due date. I'm fine with the current implementation though, won't stop me from merging.

test/TopCommandTest.py
```diff
@@ -0,0 +1,216 @@
...
+        command = TopCommand(["-s", "text", "-x", "Project1"], self.todolist, self.out, self.error)
+        command.execute()
+
+        self.assertFalse(self.todolist.is_dirty())
+        self.assertEqual(self.output, "|  3| C Baz @Context1 +Project1 key:value id:1\n|  1| C Foo @Context2 Not@Context +Project1 Not+Project\n")
+        self.assertEqual(self.errors, "")
+
+    # skip test_top20 through test_top29
+
+    # To-Do: add tests with multiline todo items, add tests of human readable dates
```
bram85: This TODO should be resolved.

test/TopCommandTest.py
```diff
@@ -0,0 +1,216 @@
...
+        config('test/data/listcommand-tags.conf')
+
+        command = TopCommand(["-s", "text", "-x", "Project1"], self.todolist, self.out, self.error)
+        command.execute()
+
+        self.assertFalse(self.todolist.is_dirty())
+        self.assertEqual(self.output, "|  3| C Baz @Context1 +Project1 key:value id:1\n|  1| C Foo @Context2 Not@Context +Project1 Not+Project\n")
+        self.assertEqual(self.errors, "")
+
+    # skip test_top20 through test_top29
```
bram85: I think this comment can be removed.

bram85:
Additionally, the commit history should be cleaned up once this PR is ready for merge. You may consider to add the Python 3.5 part and various PEP fixes in a separate PR while we iron out the `top` command in this PR.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -117,3 +154,97 @@ def filter(self, p_todo_str, _):
...
+            line3 = line2
+
+        """ Due dates """
+        pattern3 = re.compile('(?P<is_date> ' + config().tag_due() + ':(?P<year>\d{4})-(?P<month>\d{1,2})-(?P<day>\d{1,2}) ?)')
+        matches3 = pattern3.search(line3)
+        if matches3:
+            linedate = arrow.now()  # set the time to 'now' so the comparisions work better
+            linedate = linedate.replace(year=int(matches3.group('year')))
+            linedate = linedate.replace(month=int(matches3.group('month')))
+            linedate = linedate.replace(day=int(matches3.group('day')))
```
bram85: The construction of `linedate` looks a lot like the construction above. Suggestion is to put this in a separate (inline) function, such as `get_date_from_match(p_match)`.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -117,3 +154,97 @@ def filter(self, p_todo_str, _):
...
+            if addingdates is True:
+                duedelta = ', ' + duedelta
+            addingdates = True
+            line4 = pattern3.sub(' ', line3)
+        else:
+            duedelta = ''
+            line4 = line3
+
+        """ Threshold dates """
+        # To-do: implement this
```
bram85: Is this still on the radar?

bram85:
Just for my own reference, it would be nice to have [arrow issue 96](https://github.com/crsmithdev/arrow/issues/96) implemented. Then we don't see "just now" when a todo is due today, but "today".

setup.py
MinchinWeb: `arrow` 0.7 hasn't hit PyPi yet, so I'll test it once it's released there. In any case, do we want the requirements listed here to the current version, or the earliest known working version?

test/AddCommandTest.py
MinchinWeb: Most of the tests got fixed in ed2a9dd5e1f4ee2effbff962e342117f81d048d8, but this one was missed at that time. The cause would actaully be an earlier commit, 0877bac73945f765df11ff2715ecbfb99d3732a3.

test/TopCommandTest.py
MinchinWeb: Multiline tests added in a2f660a

topydo/commands/TopCommand.py
MinchinWeb: The two are fairly similar, but (almost) every function (\__init\__, _process_flags, _print, usage, help) is actually slightly different between the two.

Because they run different code, the tests for `top` are basically the same as `list`'s to make sure that if either one breaks it is caught right away.

I'm open to subclassing `ListCommand`, I'm just not sure how to do it in a way that makes sense.

topydo/lib/PrettyPrinterFilter.py
MinchinWeb: Done in 156c777b616fd5c0c3eba4e88817221bfeda7920

topydo/lib/PrettyPrinterFilter.py
MinchinWeb: Done in 44e2b9f21126da9ab7507a9e506b5872bace21f8

MinchinWeb:
@bram85 I think I have everything working here. If you're comfortable with the code, I'll work now on cleaning up the commit history.

In regards to your specific comments:
1. Indeed `arrow` could be used for [RealitiveDates](https://github.com/bram85/topydo/blob/master/topydo/lib/RelativeDate.py), but I think that is simplest dealt with as a separate issue/pull request. In particular, there is the issue of if you want to move from denoting day deltas as '1d', '3m', etc to the fuller text ('in one day', 'in one month', etc) that arrow defaults to. So I'm not going to attempt that here.
2. The dependencies have been noted in the README
3. Thanks! I'm glad you like the new formatting
4. I've implemented `top` as a separate command from `ls` because I feel it serves a different purpose. `ls` is 'list'; it serves to make my raw `todo.txt`, sort it, perhaps filter it, and make it pretty to look at. But it's output can be too long to be useful at a glance (Windows CMD defaults to a 200 line buffer, and `ls` output can easily be 500+ lines). When I want to use unfiltered `ls` output, I usually end up opening it in a text editor. `top`, on the other hand, it designed to be a quick reminder of the todo items at the top of my list, to allow me to quickly decide what to work on next (or what I've already decided to work on, but need reminding). This is why todo items are truncated at one line and output is truncated at one screenful -- this way the command's whole output can be seen as a glance. Because output is limited to ~70 characters, I'm trying to put the most important information first, and the actual 'todo item' seems more important than the dates attached to it (but I realize that could be a very subjective assessment). At an extreme, if all three dates (creation date, due date, and threshold date) are defined, they could almost fill the line themselves. Eg
```
|145| C (2 months ago, due 14 days ago, threshold of a month ago) Test item ...
```

README.md
```diff
@@ -26,6 +26,16 @@ Simply install with:
...
     pip install topydo
 
+### Required dependencies
+
+* [arrow][8]          : Used to turn dates into a human readable version.
+* [backports.shutil_get_terminal_size][9] : Used to determine your terminal
+                                            window size. This function was
+                                            added to the standard library in
+                                            Python 3.3 and so is only
+                                            required in older version of
```
bram85: Grammar fix: versions.

setup.py
bram85: It's in PyPi now.

You can leave it at 0.6 then, but please make sure nothing funny happens with the latest version (that people are likely going to install when upgrading topydo).

bram85:
Ad 1. Yes, I didn't expect this to be part of this PR, it was just an idea to create more flexibility in noting down relative dates.
Ad 2. Looks good.
Ad 4. `ls` uses a configuration option `list_limit` that defaults to `-1` (i.e. show everything). Personally I set this to 25 to prevent my console from flooding with todos. W.r.t. the dates, I think the current implementation is OK, with three dates it takes quite some space.

To elaborate a bit more on 4, I'm thinking of making the output format of `ls` (or `top`) configurable with placeholders, like the `printf` family of functions. So imagine a configuration like:

```ini
[ls]
ls_format : |%i| (%p) %s (%c, %d, %t)
list_limit: 25
```

Where `%i` is the ID, `%p` the priority, `%s` the text, `%c` the creation date, `%d` the due date, `%t` the start date. This is just a rough sketch, and by no means a request to incorporate such functionality in this PR :)

But I think such functionality would consolidate the current `ls` and `top` functionality.

topydo/lib/Utils.py
```diff
@@ -51,3 +53,26 @@ def escape_ansi(p_string):
...
+        from backports.shutil_get_terminal_size import get_terminal_size
+
+    try:
+        sz = get_terminal_size()
+    except ValueError:
+        """
+        This can result from the 'underlying buffer being detached', which
+        occurs during running the unittest on Windows (but not on Linux?)
+        """
+        Terminal_Size = namedtuple('Terminal_Size', 'columns lines')
```
bram85: Variable naming convention seems off here, suggestion to change to `terminal_size`.

setup.py
MinchinWeb: I've updated to arrow 0.7 on my local machine and everything seems to be working as excepted :)

README.md
MinchinWeb: Fixed in [c8631a](https://github.com/minchinweb/topydo/commit/c8631a6a5ec8477ea8e91295f9405e42e183a952)

topydo/lib/Utils.py
MinchinWeb: Fixed in [c8631a](https://github.com/minchinweb/topydo/commit/c8631a6a5ec8477ea8e91295f9405e42e183a952)

MinchinWeb:
I think the list_format parameter could be very powerful.

To consolidate `top` and `ls`, I think at least three additions would be needed:
 1. a parameter to truncate item output after so many lines (or maybe just one)
 2. a parameter to truncate command output after so many items
 3. ability to specify a command with argument as the default command in the configuration

This would make `top` an alias for (something like) `ls --lines 25 --single-line`


Would you rather add the parameters to `ls` rather than adding `top` as a separate command? Otherwise, I think `top` would be ready to merge in (barring fixing commit history).

bram85:
Having custom formatting and aliasing is something that is on my (mental) todo list for a while. I actually started on formatting some months ago in a local branch but I never finished it (I think it's impossible to merge now anyway).

I was thinking recently of adding a `-n` flag to `ls` to limit the number of commands (it's similar to head(1) and tail(1)). I thought I actually implemented it, but apparently I didn't :)

I think I can merge in `top` for now, and replace it once the formatting/aliasing/limiting is in place. The tests will make sure that nothing of its behavior changes at that point.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -125,3 +158,115 @@ def filter(self, p_todo_str, _):
...
+        #   |h2j| (C) ...
+        if matches:
+            if matches.group('id'):
+                if matches.group('pri'):
+                    """ If we have a priority """
+                    return(matches.group('id') + matches.group('pri') + ' ' + p_todo_str[matches.span()[1]:])
+                else:
+                    return(matches.group('id') + self.p_replacement + ' ' + p_todo_str[matches.span()[1]:])
+            else:
+                if matches.group('pri'):
```
bram85: I ran `coverage`, it seems you missed this case in the tests.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -125,3 +158,115 @@ def filter(self, p_todo_str, _):
...
+                    """ If we have a priority """
+                    return(matches.group('pri') + ' ' + p_todo_str[matches.span()[1]:])
+                else:
+                    return(self.p_replacement + ' ' + p_todo_str[matches.span()[1]:])
+        else:
+            """ No match found """
+            return(self.p_replacement + ' ' + p_todo_str)
```
bram85: This case was also missed according to `coverage`.

bram85:
I didn't have the opportunity yet to run `coverage` on your branch, I was able to do so as you can see. Sorry to add these comments in the nearly-ready-for-merge state.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -125,3 +158,115 @@ def filter(self, p_todo_str, _):
...
+        Takes a match group, compares it to 'now', and then returns a 'human
+        readable' version of the difference.
+
+        Assumes the matchgroup has matches named 'year', 'month', and 'day'.
+        """
+        linedate = arrow.now()  # set the time to 'now' so the comparisions work better
+        linedate = linedate.replace(year=int(matchgroup.group('year')))
+        linedate = linedate.replace(month=int(matchgroup.group('month')))
+        linedate = linedate.replace(day=int(matchgroup.group('day')))
+        delta = linedate.humanize()
```
bram85: Suggestion is to call:

```python
delta = linedate.humanize(only_distance=True)
```

and do the grammar ourselves. Using this parameter requires us to bump the arrow version to 0.7.0.

Rationale: `(7 days ago, due in 8 days, threshold of in 4 days)` looks funny.

What do you think?

MinchinWeb: yes, `threshold of in 4 days` reads funny. I'll look into it...

MinchinWeb: Fixed in 0130560. New output:

```
|354| Z +test (just now, due just now, threshold of 9 months ago)
|353| Z +test (just now, threshold in a year)
```

I think the grammar on due and added dates is working as arrow provides, at least in English. If the program is translated, that could bring up other issues.

topydo/lib/PrettyPrinterFilter.py
MinchinWeb: Tested added in 2e915bb.

bram85:
FYI, in my branch `ls-n` I introduced a `-n` flag to `ls`:

`topydo ls -n 5`

That is one step towards making `ls` more flexible as outlined in a previous comment. I will merge it once this branch has been merged, because it conflicts with your branch (test cases fail).

mruwek:
I've finally had some time to try this PR at home. Couple of comments from my perspective as a user:

1. I would like to **always** have an *option* (or collection of options) to get output of `topydo ls` which would be "compatible" with the output of original `todo.sh ls`. So with this change brackets surrounding priority are gone - I would like to have an *option* to get them back. Maybe we can include removing brackets to that already introduced option and rephrase it as `human_readable_format`? Of course this comment will lose its meaning when/if `list_format` will be implemented.
2. `topydo ls -h` still feels more natural than `topydo ls -r`. I know that there is already `topydo -h` for general help, but: they appear in other option chains (general option vs subcommand option). If we can't have two meanings for `-h` let's get rid of `-h` for help, after all there is `topydo help`. Other apps very rarely use anything other than `-h` with *human-readable-format* in mind (consider `du`, `df` and... `ls` in unix world)
3. Human readable dates should have the same color as "raw" dates in my opinion.

One remark from occasional coder perspective: I believe we could include python-dateutil in README.md required dependencies as arrow depends on it. It's indirect dependency, so it's not a must for us, but could be helpful information.

All in all, it's really nice improvement

mruwek:
One more thing. I think boolean defaults should be contrary to values of lonely command-line flags. So, if we do have command-line switch (flag) for `human_readable_dates` defined as `True`, the default value should be `False`. It's quite strange for me to have lonely cli-flag for some option and meaning of this flag is the same as default value. If we would have both flags (`True` and `False`) in cli it's not a problem at all, but when we do have only one of them, default value should be contrary to it.

bram85:
Thanks for your remarks, @mruwek .

* After using this PR for a couple of working days, I also feel there's a mismatch with the configuration option and the flag. There are rare occasions that I would like to see the absolute dates, but that's quite hard to accomplish in the current implementation (you can only temporarily set `human_readable_dates` to `0` in the config file).

  I do think one flag is sufficient, you only want to occasionally disable relative dates, not occasionally enable them. We can keep the `-r` flag, but change it's meaning to show **r**aw dates when you have configured topydo to show relative dates. I.e. reverse the meaning. That also eliminates `-h` vs. `-r`.

* I have no strong feelings for `(C)` vs `C`. I have no problems with the current format in this PR. `human_readable_format` that affects priority + relative dates is a suggestion that you can consider.

* Colorization of the dates is a good suggestion, IMHO.

All these suggestions aside, I enjoy working with the nice(r) formatting of `ls`.

mruwek:
> We can keep the -r flag, but change it's meaning to show raw dates when you have configured topydo to show relative dates. I.e. reverse the meaning. That also eliminates -h vs. -r.

+1 for this.

>  I have no problems with the current format in this PR.

As I said - output "compatible" with original `todo.sh` would be a nice option to have and best solution for that will surely be `list_format`. Any ETA on that? :)

MinchinWeb:
@bram85 @mruwek 
- I have switched around the behavior of the `-r`command line flag so that it provides 'raw dates' and overrides whatever the 'human readable dates' setting in the configuration is.
- I've added `python-dateutil` to the README

- > '3. Human readable dates should have the same color as "raw" dates in my opinion.

  I'm not sure I understand what you are suggesting. Can you explain it more?

- > All these suggestions aside, I enjoy working with the nice(r) formatting of `ls`.

   I'm so glad :)


I think the last thing left is to work on the history. Reworking this is surprisingly daunting. Do you have any suggestions, or read any good articles about how to go about it? What do you want accomplished at the end?

bram85:
> - > '3. Human readable dates should have the same color as "raw" dates in my opinion.

>  I'm not sure I understand what you are suggesting. Can you explain it more?

To color the dates with the metatag color, if configured (see [here](https://github.com/bram85/topydo/blob/stable/topydo/lib/PrettyPrinterFilter.py#L74)).

However, when looking at it, I don't like the color functionality to be spread out over multiple filters (I prefer it to be local to the `PrettyPrinterColorFilter`.

I think this issue can (also) be addressed with `ls --format` functionality, to have color codes and let the user decide what to color (and leave the colors of the dates as-is at this moment).

MinchinWeb:
Ok, I'll leave colored dates alone for now.

test/TestConfig.py
```diff
@@ -15,28 +15,135 @@
...
+        self.assertEqual(config("test/data/ConfigTest1.conf",
+                                p_overrides=overrides).default_command(),
+                         'edit')
+
+    def test_config05(self):
+        """ Bad colour switch value. """
+        # boolean settings must first be typecast to integers, because all
+        #  strings evaulate to 'True'
+        self.assertEqual(config("test/data/ConfigTest4.conf").colors(),
+                         bool(int(config().defaults["colors"])))
```
bram85: I think these tests need to be fixed after the inclusion of PR #58 . Better is perhaps to have a dedicated PR for this as well, since it's not strictly related to the `top` command.

bram85:
> I think the last thing left is to work on the history. Reworking this is surprisingly daunting. Do you have any suggestions, or read any good articles about how to go about it?

Not aware of any good articles on how to deal with this. I assume you're familiar with `git rebase -i origin/master` ?

Ideally I would like to see "just enough" commits. Bugfixes that we encountered along the way should not be visible in the PR history, they should be squashed with the commit that introduced the new functionality. Fixes on .gitignore should be in a separate commit (I believe it already is).

Hopefully it helps.

mruwek:
> I think the last thing left is to work on the history. Reworking this is surprisingly daunting. Do you have any suggestions (…)

`git rebase -i master` is the key. Sometimes it's as easy as reordering + squashing relevant commits. Sometimes it's more problematic.

Before merging #47 I've done something like this:

`git checkout -b yourbranch-rebase` - let's make squashing in another "safe" branch
`git rebase -i master` - squash everything into **one** really big commit - save whole commit message for easier copy-pasting parts of it later.
`git reset HEAD~` - this will break all your changes into **unstaged** state
`git add` & `git commit` to produce as many commits as you want or `git add --patch` to do this interactively.

bram85:
> To elaborate a bit more on 4, I'm thinking of making the output format of ls (or top) configurable with placeholders

FYI, I actually started working on this in my [ls-format branch](https://github.com/bram85/topydo/tree/ls-format).

Feedback, PRs, etc. are welcome :)

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -125,3 +157,120 @@ def filter(self, p_todo_str, _):
...
+        """
+        Takes a match group and then returns a date.
+
+        Assumes the matchgroup has matches named 'year', 'month', and 'day'.
+
+        The returned arrow object has hours, minutes, and seconds set to now.
+        """
+        linedate = arrow.now()  # set the time to 'now' so the comparisions work better
+        linedate = linedate.replace(year=int(matchgroup.group('year')))
+        linedate = linedate.replace(month=int(matchgroup.group('month')))
```
bram85: There's a bug here, by replacing the items one by one.

The bug is that this sequence may cause an invalid (intermediate) date, causing topydo to crash.

* Today is October 30th 2015 (2015-10-30).
* I have an item with creation date 2015-02-09.
* At this point in the code, a date of 2015-02-30 is created, which is an invalid date. Crash.

 Suggestion is to replace them in once call, if that works. Or, if you're sure what you're doing, catch `ValueError`.

Also make sure that this is covered as a test case.

MinchinWeb: Good catch! Fixed in 49b3cf0b0b2aadcbb47a953eac6cb069e55ec525

PR #53, opened 2015-09-27
Much more powerful filtering of priorities

mruwek:
Now it is possible to do things like:
`ls (>C)` # gives items with D-Z priorities
`ls (!E)` # gives all items except those with E priority (also possible with `ls -- -(E)`)

and maybe most importantly:
`ls (<A)` # gives items without priority

If You like it, I'll complete this pull-request with some tests and docstrings of course.

PS Don't forget about escaping parentheses when calling from CLI

### DIFF
--- a/topydo/lib/Filter.py
+++ b/topydo/lib/Filter.py
@@ -151,19 +151,49 @@ class LimitFilter(Filter):
     def filter(self, p_todos):
         return p_todos[:self.limit] if self.limit >= 0 else p_todos
 
-ORDINAL_TAG_MATCH = r"(?P<key>[^:]*):(?P<operator><=?|=|>=?|!)?(?P<value>\S+)"
-
-class OrdinalTagFilter(Filter):
-    def __init__(self, p_expression):
-        super(OrdinalTagFilter, self).__init__()
+class OrdinalFilter(Filter):
+    """
+    Base class for ordinal filters.
+    """
+    def __init__(self, p_expression, p_pattern):
+        super(OrdinalFilter, self).__init__()
 
         self.expression = p_expression
 
-        match = re.match(ORDINAL_TAG_MATCH, self.expression)
+        match = re.match(p_pattern, self.expression)
         if match:
-            self.key = match.group('key')
+            try:
+                self.key = match.group('key')
+            except IndexError:
+                pass
             self.operator = match.group('operator') or '='
             self.value = match.group('value')
+
+    def compare_operands(self, p_operand1, p_operand2):
+        """
+        Returns True if conditional constructed from both operands and
+        self.operator is valid. Returns False otherwise.
+        """
+        if self.operator == '<':
+            return p_operand1 < p_operand2
+        elif self.operator == '<=':
+            return p_operand1 <= p_operand2
+        elif self.operator == '=':
+            return p_operand1 == p_operand2
+        elif self.operator == '>=':
+            return p_operand1 >= p_operand2
+        elif self.operator == '>':
+            return p_operand1 > p_operand2
+        elif self.operator == '!':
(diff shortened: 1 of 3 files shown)


bram85:
I like the idea, thank you. So I'll merge it once the feature is complete.
I'll put some comments in the code.

topydo/lib/Filter.py
```diff
@@ -214,3 +214,35 @@ def match(self, p_todo):
...
+        match = re.match(PRIORITY_MATCH, self.expression)
+        if match:
+            self.operator = match.group('operator') or '='
+            self.priority = match.group('priority')
+
+    def match(self, p_todo):
+        operand1 = p_todo.priority()
+        operand2 = self.priority
+
+        if self.operator == '<':
```
bram85: This piece is identical to the code in `OrdinalTagFilter`, perhaps this part can be reused in a superclass, e.g. `OrdinalFilter`, with subclasses `OrdinalTagFilter` and `OrdinalPriorityFilter` (or just `PriorityFilter` as you wrote). The subclasses take care of setting the operands, while the base class evaluates the operator and the operands.

Because I can think of a future extension to filter on creation date, which would use the same functionality, only a different attribute to extract from the todo.

topydo/lib/Filter.py
```diff
@@ -214,3 +214,35 @@ def match(self, p_todo):
 
         return False
 
+PRIORITY_MATCH = r"\((?P<operator><=?|>=?|=|!)?(?P<priority>[A-Z]{1})\)"
```
bram85: The first part can be put in a separate regexp, e.g. `OPERATOR_MATCH`, that can be used by this match expression and the `ORDINAL_TAG_MATCH`.

topydo/lib/Filter.py
mruwek: done in cd41de0

topydo/lib/Filter.py
mruwek: done in fa9398e

topydo/lib/Filter.py
bram85: Nice, this is what I had in mind.

If you add the tests + docstrings I'll merge it in.

bram85:
Looking at the test cases I feel that the semantics of the comparison operator are not so intuitive for me.

In my head I formulate a query "give me all priorities *higher* than B", which in this implementation would be written down as `(<B)`.

Moreover, I consider todos without priority to be lower than Z, while in your ordering they're "lower than A". The first semantics is also how the `sort` command treats todos without priority (see [Sorter.py:38](https://github.com/bram85/topydo/blob/stable/topydo/lib/Sorter.py#L38))

What do you think?

mruwek:
> What do you think?

I agree with You. I've implemented it that way because of python method of comparing strings (`'A' > 'B' ` returns `False`). It won't be hard however to change it, so expect rebase in a moment :)

mruwek:
> In my head I formulate a query "give me all priorities higher than B", which in this implementation would be written down as (<B).

Ok. Now `ls (>B)` will give items with priority A.

> Moreover, I consider todos without priority to be lower than Z, while in your ordering they're "lower than A".

Fixed in 2c5b7175c7bf3b10e0732d3ab9be9bf6eb271ca8. I needed dummy value when `Todo.Todo.priority()` returns `None`. At first I've used "^", but then I've noticed that You used "ZZ" in sorter.py for same purpose as mine.

mruwek:
I've also reviewed commit message for 2c5b7175c7bf3b10e0732d3ab9be9bf6eb271ca8 and doscstring in `Filter.PriorityFilter.match()`.

bram85:
Looks great, thank you.

PR #56, opened 2015-10-12
Fix Wheel building

MinchinWeb:
Wheels downloaded from PyPI weren't installing `colorama` for me on WIndows (colorama is specified as a required dependency, but only for Windows).

I believe this fixes it by moving the evaluation of conditional requirements at install time rather than compile time.

`python setup.py install` and `pip install .` (from the project root folder) both still work.

See [Wheel documentation](https://wheel.readthedocs.org/en/latest/index.html#defining-conditional-dependencies) for more details.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,11 +21,6 @@ def find_version(*file_paths):
     raise RuntimeError("Unable to find version string.")
 
 
-conditional_dependencies = {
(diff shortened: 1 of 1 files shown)


bram85:
Thank you.

MinchinWeb:
Glad to help :)

PR #57, opened 2015-10-21
Aliases

mruwek:
Aliases can now be defined in 'aliases' section of topydo config file.

Examples:

```
[aliases]
showall = ls -x
purge = del -f
ical = ls -f ical
```


MinchinWeb:
Can you use an alias as the default command?

mruwek:
@MinchinWeb
> Can you use an alias as the default command?

Now You can (after pushing d092693).

topydo/lib/Config.py
```diff
@@ -266,6 +267,22 @@ def auto_creation_date(self):
         except ValueError:
             return self.defaults['auto_creation_date'] == '1'
 
+    def aliases(self):
+        """
+        Returns dict with aliases names as keys and pairs of actual
+        subcommand and alias args as values.
+        """
+        aliases = list(set(self.cp.items('aliases')) - set(self.cp.defaults().items()))
```
bram85: * What is the purpose of this line exactly?
* Please add a comment with a clarification.
* Does it have to be a `list` ? Why not leave it as a set and iterate over that below? I don't think the order matters anymore after turning the configuration items into sets.

topydo.conf
```diff
@@ -46,3 +46,6 @@ append_parent_contexts      = 0
 ; context_color = magenta
 ; metadata_color = green
 ; link_color = light-cyan
+
+[aliases]
+;showall = ls -x
```
bram85: My suggestion is to add all variants of `lsprj` and `lscon` as a "pre-installed" alias (and remove the variants from the `_SUBCOMMAND_MAP`).

topydo/Commands.py
```diff
@@ -52,6 +52,7 @@
     'sort': 'SortCommand',
     'tag': 'TagCommand',
 }
+_alias_map = config().aliases()
```
bram85: Naming convention: `_ALIAS_MAP`.

bram85:
Very nice, I'll give it a try locally.

mruwek:
Updated and rebased to support new default config model introduced with #58.

bram85:
Looks good. I think we overlooked one thing: auto-completion in prompt mode. Shouldn't be hard to merge the normal subcommands and the aliases in the `_subcommands` function in TopydoCompleter.py.

PR #63, opened 2015-10-31
Properly split alias options wrapped in quotes

mruwek:
Fixed with `shlex`.


bram85:
Can we cover this with one or more test cases?

mruwek:
Of course. I simply didn't have enough time to write them. I'll do them tonight.

PR #64, opened 2015-11-01
Add customizable right alignment in list_format

mruwek:
User can now mark place in list_format from where he wants to start
right-aligning remaining content. Tab character (\t) is used as marker.

Added also 'top' alias example to topydo.conf which utilizes this
feature and implements #52.

Minor: remove `- 1` from statements setting line width.

### DIFF
--- a/topydo/lib/ListFormat.py
+++ b/topydo/lib/ListFormat.py
@@ -52,3 +52,36 @@ def humanize_dates(p_due=None, p_start=None, p_creation=None):
             dates_list.append('starts in ' + start)
 
     return ', '.join(dates_list)
+
+def strip_placeholder_braces(p_matchobj):
+    """
+    Returns string with conditional braces around placeholder stripped and
+    percent sign glued into placeholder character.
+
+    Returned string is composed from 'start', 'before', 'placeholder', 'after',
+    'whitespace', and 'end' match-groups of p_matchobj. Conditional braces are
+    stripped from 'before' and 'after' groups. 'whitespace', 'start', and 'end'
+    groups are preserved without any change.
+
+    Using this function as an 'repl' argument in re.sub it is possible to turn:
+        %{(}B{)}
+    into:
+        (%B)
+    """
+    try:
+        before = p_matchobj.group('before').strip('{}')
+    except AttributeError:
+        before = ''
+
+    placeholder = p_matchobj.group('placeholder')
+
+    try:
+        after = p_matchobj.group('after').strip('{}')
+    except AttributeError:
+        after = ''
+
+    whitespace = p_matchobj.group('whitespace') or ''
+    start = p_matchobj.group('start') or ''
+    end = p_matchobj.group('end') or ''
+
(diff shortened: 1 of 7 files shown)


topydo/lib/PrettyPrinterFilter.py
```diff
@@ -233,10 +233,32 @@ def strip_braces(p_matchobj):
...
-                        line_width = get_terminal_size().columns -1
+                        line_width = get_terminal_size().columns
                         if len(p_todo_str) > line_width:
-                            text_lim = line_width - len(p_todo_str) - 3
+                            text_lim = line_width - len(p_todo_str) - 4
                             p_todo_str = re.sub(re.escape(repl), repl[:text_lim] + '...', p_todo_str)
 
         return p_todo_str
 
+class PrettyPrinterBeautyFilter(PrettyPrinterFilter):
```
bram85: `PrettyPrinterAlignFilter` would be a better name.

MinchinWeb:
@mruwek, the reason for capping printed lines at one less than the console width is (on Windows, at least) that if you print a line that is exactly the line's width, the next line will be blank.

topydo/lib/ListFormat.py
```diff
@@ -52,3 +52,22 @@ def humanize_dates(p_due=None, p_start=None, p_creation=None):
             dates_list.append('starts in ' + start)
 
     return ', '.join(dates_list)
+
+def strip_ph_braces(p_matchobj):
```
bram85: Readability: Write out `ph` to `placeholder`.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -205,41 +204,47 @@ def filter(self, p_todo_str, p_todo):
...
+    Currently it only provides right alignment from place specified in
+    list_format config-option (subsitutes tab-character with as many spaces as
+    it is needed to fill the whole line).
+    """
 
+    def __init__(self):
+        super(PrettyPrinterAlignFilter, self).__init__()
+
+    def filter(self, p_todo_str, _):
+        line_width = get_terminal_size().columns
```
bram85: Variable scope: `line_width` and `to_fill` can be calculated under `if tab`.

topydo/lib/PrettyPrinterFilter.py
```diff
@@ -205,41 +204,47 @@ def filter(self, p_todo_str, p_todo):
...
+        to_fill = line_width - len(p_todo_str)
+        tab = re.match('.*\t', p_todo_str)
```
bram85: Instead you could also use `re.search('\t', p_todo_str)`.

mruwek:
@MinchinWeb Take note I'm removing **4** characters and adding **3** ('...') when calculating `%S` so final result is always one character shorter than terminal width. I've tested that on CMD.EXE under Win7 with various terminal sizes and everything worked as expected.

MinchinWeb:
@mruwek Sounds good. I guess I caught a comment out of context. Thanks for the explanation.

PR #65, opened 2015-11-03
Fix help in prompt mode

mruwek:
`help` is not a real subcommand and without arguments it is of NoneType,
so we need to properly handle it.


bram85:
Thank you for the fix.

PR #68, opened 2015-11-05
A couple more tests

MinchinWeb:
- add more tests of the configuration
- add a test that covers the `help` subcommand for the `edit` command
- allow revert tests to create tempfiles in a cross-platform way


bram85:
Thanks, looks good except for the skipped test cases. I prefer not to fill topydo with test cases that are not executed, which will "eventually" be implemented.

bram85:
Thank you.

PR #72, opened 2015-11-09
Update tests for ls_format

mruwek:


mruwek:
Sorry - opened against master instead of ls-format.

PR #76, opened 2015-11-10
Slightly modify grammar of ls_format

mruwek:
- From now on literal `%` signs always have to be escaped with `\` (`\%`),
even inside conditional braces.
- `%%` won't work anymore.
- Nonexistent placeholders (for example`%&`) will be expanded to empty
  strings (`''`).

This change fixes also some bugs:
- inappropriate output width when \t was used without %S
  (example: `%p{ }\t%{ }p`)
- curly braces inside conditional braces (example: `%{{}p{}}`)

### DIFF
--- a/topydo/lib/ListFormat.py
+++ b/topydo/lib/ListFormat.py
@@ -68,20 +68,9 @@ def strip_placeholder_braces(p_matchobj):
     into:
         (%B)
     """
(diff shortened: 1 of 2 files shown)


bram85:
Thanks a lot for the rework!

PR #77, opened 2015-11-13
Move format parsing logic to ListFormatParser

mruwek:
Main advantage of having separate parser for list_format is possibility of initializing it and resolving placeholders only once. Previous implementation did that with each Todo object. `ListFormatParser` also resolves placeholders differently - it doesn't iterate through a dictionary with placeholders and look for a match anymore. Now it finds the right placeholder and simply retrieve it from dictionary.

I'm also thinking about moving `PrettyPrinterAlignFilter` logic into that class.

### DIFF
--- a/test/test_list_format.py
+++ b/test/test_list_format.py
@@ -631,5 +631,22 @@ Z   Z
 """
         self.assertEqual(self.output, result)
 
+    @mock.patch('topydo.lib.ListFormat.get_terminal_size')
+    def test_list_format42(self, mock_terminal_size):
+        mock_terminal_size.return_value = self.terminal_size(100, 25)
+
+        config('test/data/listformat.conf', p_overrides={('ls', 'indent'): '3'})
+        command = ListCommand(["-x"], self.todolist, self.out, self.error)
+        command.execute()
+
+        result = u"""   |  1| (D) 2015-08-31 Bar @Context1 +Project2                         due:2015-09-30 t:2015-09-29
+   |  2| (Z) 2015-11-06 Lorem ipsum dolorem sit amet. Red @... due:2015-11-08 lazy:bar t:2015-11-07
+   |  3| (C) 2015-07-12 Foo @Context2 Not@Context +Project1 Not+Project
+   |  4| (C) Baz @Context1 +Project1                                                      key:value
+   |  5| Drink beer @ home                                                     ical:foobar id:1 p:2
+   |  6| x 2014-12-12 Completed but with                                            date:2014-12-12
+"""
+        self.assertEqual(self.output, result)
+
 if __name__ == '__main__':
     unittest.main()
--- a/test/test_list_format.py
+++ b/test/test_list_format.py
@@ -52,7 +52,7 @@ class ListFormatTest(CommandTest):
 """
         self.assertEqual(self.output, result)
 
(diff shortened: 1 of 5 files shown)


test/test_list_format.py
```diff
@@ -598,5 +598,38 @@ def test_list_format39(self):
 """
         self.assertEqual(self.output, result)
 
+    @mock.patch('topydo.lib.ListFormat.get_terminal_size')
+    def test_list_format40(self, mock_terminal_size):
```
bram85: What is this test about?

topydo/lib/ListFormat.py
```diff
@@ -74,3 +85,146 @@ def strip_placeholder_braces(p_matchobj):
...
+
+    return clean_str
+
+class ListFormatParser(object):
+    def __init__(self, p_todolist, p_format=None):
+        self.format_string = re.sub(r'\\t', '\t', p_format or config().list_format())
+        self.todolist = p_todolist
+        self.one_line = False
+        self.line_width = get_terminal_size().columns
+        self.placeholders = {
```
bram85: Can be moved outside the class.

topydo/commands/ListCommand.py
```diff
@@ -88,11 +88,10 @@ def _print(self):
             # create a standard printer with some filters
             indent = config().list_indent()
             hidden_tags = config().hidden_tags()
+            format_parser = ListFormatParser(self.todolist, self.format)
 
             filters = []
-            filters.append(PrettyPrinterFormatFilter(self.todolist,
-                           self.format))
-            filters.append(PrettyPrinterAlignFilter())
+            filters.append(PrettyPrinterFormatFilter(self.todolist, format_parser))
```
bram85: I would like to see that `PrettyPrinterFormatFilter` instantiates the parser, rather than putting that responsibility to the client of the filter.

topydo/lib/ListFormat.py
```diff
@@ -74,3 +85,146 @@ def strip_placeholder_braces(p_matchobj):
     whitespace = p_matchobj.group('whitespace') or ''
 
     return before + '%' + placeholder + after + whitespace
+
+def str_clean(p_str):
```
bram85: This function does two things, it should be split, with their names describing what it does.

topydo/lib/ListFormat.py
mruwek: Moving placeholders dict outside of the class will create unnecessary problem with `%i` and `%I` in my opinion, as they depend on `Todolist` object.

test/test_list_format.py
mruwek: `test_list_format40` tests if specifying `\t` in `list_format` option works. Relevant code part is in [`ListFormatParser.__init__()`](https://github.com/bram85/topydo/pull/77/files#diff-c0cff3e2ccdd829757f686fe501da792R97).

`test_list_format41` tests `\t` specified from CLI as an argument for `-F` flag.

topydo/commands/ListCommand.py
mruwek: fixed in b72351e

topydo/lib/ListFormat.py
mruwek: Splitted in: b72351e

bram85:
Change looks good overall, but please add docstrings.

PR #83, opened 2015-11-30
Alternate for `top` command

MinchinWeb:
The list formatting options combined with aliases make a good combination for providing the `top` command (originally proposed in PR #52) with less duplicated code. This pull request adds to 'mini-features' to provide the `top` command like I'd imagined.

- a placeholder space if no priority is specified (`%P`)
- an option to just print a screenful of lines (`ls -N`)

I've also updated the `top` alias in the default configuration to what I'm using locally.

Closes #52.


topydo/commands/ListCommand.py
```diff
@@ -193,6 +198,8 @@ def help(self):
 
      A tab character serves as a marker to start right alignment.
 -i : Comma separated list of todo IDs to print.
+-n : Number of items to display. Defualts to the value in the configuration.
```
bram85: Small typo here: _Defualts_

topydo/commands/ListCommand.py
```diff
@@ -193,6 +198,8 @@ def help(self):
...
+-N : Display one 'screenful' of items.
```
bram85: I'd rephrase it to _Limit number of items such that it fits on the terminal_.

topydo/commands/ListCommand.py
```diff
@@ -69,11 +70,14 @@ def _process_flags(self):
                     self.printer = None
             elif opt == '-F':
                 self.format = value
+            elif opt == '-N':
+                # 2 lines are assumed to be taken up by printing the next prompt
+                self.limit = get_terminal_size().lines - 2
```
bram85: What if `get_terminal_size().lines() < 2`? I think the value should be `max()`'ed to 0.
I'd like to see this covered in a test case.

bram85:
Change looks good. Some remarks:

* Tests are missing for the `-N` flag and the `%p` placeholder.
* I'm in doubt whether `%p` (lowercase) should expand to a space if empty, instead of a new placeholder. But I tend to say that your proposal is better to have a separate placeholder.

topydo/commands/ListCommand.py
MinchinWeb: Fixed in 28fbec77a08f9dc677d73f4fb01666b7525436d3.

topydo/commands/ListCommand.py
MinchinWeb: Done in 9b766e9d4633055efc815c9365c6048c210294a7.

topydo/commands/ListCommand.py
MinchinWeb: Done in 9b766e9d4633055efc815c9365c6048c210294a7.

I actually set it up to print at least one item; that seemed more practical.

MinchinWeb:
This is actually set up with two parameters:

- `%p` (lowercase) is left unchanged -- this displays nothing if there is no priority
- `%P` (uppercase) is new -- prints a space if there is no priority

I'm fine if they are combined; i just added a new one because I didn't want to take away any functionality that someone else might be assuming is still there.

I've added test for both the `-N` flag and the `%P` (uppercase) placeholder. I also added to code to fix import issues in PyPy (which is based on Python 3.2).

bram85:
How come we missed those dependencies on Travis CI? E.g. build [94735116](https://travis-ci.org/bram85/topydo/jobs/94735116) went fine.

MinchinWeb:
I think in mocking the `topydo.Utils.get_terminal_size()` throughout the test suite, this function was never really called directly. The only reason I caught it was that I moved the imports (of `shutil.get_terminal_size()` and `backports.shutil_get_terminal_size.get_terminal_size()`) from the `topydo.Utils.get_terminal_size()` function to the top of the ``topydo/Utils.py`file and then `pylint` started complaining.

We probably missed `mock` as it is installed by default on Travis-CI ([list of preinstalled packages](https://docs.travis-ci.com/user/languages/python#Pre-installed-packages)).

bram85:
Thank you. Only thing missing wrt #52 is the indenting of multi-line items. I got the impression from @mruwek that it was not a trivial thing to do.

MinchinWeb:
Thanks for merging this!

The other thing missing is updating the wiki.

I can look into the multi-line formatting. It seem like it shouldn't be too complicated, but I guess I'll find out how naive I am shortly...

mruwek:
@MinchinWeb 
Multi-line formatting consists (briefly) of:
1. text-wrapping
2. indenting second and following lines of one todo items

Text-wrapping should be easy (as you can reuse to great extent your own code from #52). I'm concerned only about indentation. I simply don't know how deep we should indent second and following lines and if we should make this a fixed value or configurable one or calculable one. Implementation part should be pretty straightforward after making those decisions :)

MinchinWeb:
@mruwek the `textwrap` module I was using in #52 deals with indenting following lines quite simply, so that's not a worry. The question is, How much to indent?

What about if you were to base it on the format string. Divide format parameters into 'prefix' and 'body' categories. Any 'prefix' parameters before the first 'body' parameter increase the indent. So, if your format string is `%I %x %P %s %k %{(}H{)}`:

base indent : +2
%I : +3
(space) : +1
%x : +1
(space) : +1
%P : +1
(space) : ignored because next item is a body parameter
%s : body parameter, we're done

So we get an indent of [2+3+1+1+1+1 = ] 9.

Does this seem like a workable solution?

bram85:
Or a new placeholder that serves as a marker for example `%>`:

`%I %x %P %>%s %k %{(}H{)}`

More flexible, no need to categorize placeholders or calculate the length based on the body.

What do you think?

MinchinWeb:
Yes, I like it. I think it would work well for the vast majority of cases.

So something like:
`%I %x %P %s %k %{(}H{)}` -- (no marker, so) no second line indent
`%> %I %x %P %s %k %{(}H{)}` -- 2 space second line indent (marker at front of line)
`%>%> %I %x %P %s %k %{(}H{)}` -- 4 space second line indent (double marker at front of line)
`%I %x %P %> %s %k %{(}H{)}`-- 9 space second line indent (as per previous comment)
`%I %x %P %>%> %s %k %{(}H{)}`-- 11 space second line indent (as per previous comment)

PR #84, opened 2015-11-30
Priority command: allow lowercase priorities to be input

MinchinWeb:
Allow both upper- and lowercase priorities to be specified for the `pri` command. So these two commands (now) give the same results:

```
> topydo pri 2 D
> topydo pri 2 d
```

### DIFF
--- a/topydo/commands/PriorityCommand.py
+++ b/topydo/commands/PriorityCommand.py
@@ -22,7 +22,7 @@ from topydo.lib.Utils import is_valid_priority
 
 
 class PriorityCommand(MultiCommand):
-    def __init__(self, p_args, p_todolist, #pragma: no branch
+    def __init__(self, p_args, p_todolist,  # pragma: no branch
                  p_out=lambda a: None,
                  p_err=lambda a: None,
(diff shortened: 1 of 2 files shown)


topydo/commands/PriorityCommand.py
```diff
@@ -36,7 +36,7 @@ def normalize_priority(p_priority):
             match = re.search(r'\b([A-Z])\b', p_priority)
             return match.group(1) if match else p_priority
 
-        priority = normalize_priority(self.args[-1])
+        priority = normalize_priority(self.args[-1].upper())
```
bram85: I see it as a part of normalization, so I'd expect `upper()` to be called in the `normalize_priority` function.

```python
def normalize_priority(p_priority):
    match = re.search(r'\b([A-Z])\b', p_priority.upper())
    return match.group(1) if match else p_priority
```

MinchinWeb: Good call. Changed in 9b855ec.

MinchinWeb:
Rebased and updated as per @bram85 's comments.

PR #85, opened 2015-12-16
Create placeholder for args in aliases

mruwek:
With this change there is now possible to forward args supplied on CLI
to desired placed in real command.

Example config:

```
[aliases]
star = tag {} star 1
unstar = tag {} star
```

Example usage:

`topydo star foo` will resolve to: `topydo tag foo star 1`
`topydo unstar foo` will resolve to: `topydo tag foo star`


bram85:
Thank you, looks good.

PR #87, opened 2015-12-26
Introduce config file for column layout

mruwek:
Column-ui will look for columns configuration in:

1. system-wide configuration:
/etc/topydo_columns.conf

2. user dir:
~/.topydo_columns
~/.config/topydo/columns

3. current dir:
.topydo_columns
topydo_columns.conf
topydo_columns.ini

Each column should have its own INI style config section:

```ini
[FooBar]
title = Foo
filterexpr = Bar
sortexpr = desc:prio
show_all = 1
```

If any column option is missing, default value will be used.


mruwek:
Bad target branch. Sorry.

PR #90, opened 2015-12-26
Add `topydo columns` subcommand to launch column-ui

mruwek:


topydo/cli/UILoader.py
```diff
@@ -42,6 +42,12 @@ def main():
                 PromptApplication().run()
             except ImportError:
                 error("You have to install prompt-toolkit to run prompt mode.")
+        elif args[0] == 'columns':
+            try:
+                from topydo.ui.Main import UIApplication
+                UIApplication().run()
+            except ImportError:
+                error("You have to install urwid to run column UI.")
         else:
```
bram85: Change UI to mode, then it's more consistent with prompt mode (rather than prompt UI).

mruwek: Done in 669e062

PR #97, opened 2015-12-31
Add history to command-line widget

mruwek:
Main concept and great part of implementation derived from
[pazz/alot][1].

[1]: https://github.com/pazz/alot/blob/0.3.7/alot/widgets/globals.py#L76

### DIFF
--- a/topydo/ui/CommandLineWidget.py
+++ b/topydo/ui/CommandLineWidget.py
@@ -18,6 +18,12 @@ import urwid
 
 class CommandLineWidget(urwid.Edit):
     def __init__(self, *args, **kwargs):
+
+        self.history = None
+        self.history_pos = None
+        # temporary history storage for edits before cmd execution
+        self.history_tmp = None
+
         super().__init__(*args, **kwargs)
         urwid.register_signal(CommandLineWidget, ['blur', 'execute_command'])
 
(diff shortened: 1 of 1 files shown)


topydo/ui/CommandLineWidget.py
```diff
@@ -30,12 +35,36 @@ def _blur(self):
...
+                self.history = []
+            self.history.append(self.edit_text)
+        self.history_pos = len(self.history)
+        self.history_tmp = self.history[:]
+        self.history_tmp.append('')
+
+    def _history_move(self, p_step):
+        if self.history is not None:
+            self.history_tmp[self.history_pos] = self.edit_text
+            self.history_pos = (self.history_pos + p_step) % len(self.history_tmp)
```
bram85: I don't think that `self.history_pos + p_step` should ever evaluate to a negative value, that would happen when pressing down while not having moved up. You'd suddenly see the oldest entry in history, which unexpected (at least for me). Is this the behavior in prompt mode (or bash/zsh)?

mruwek: You're right. Fixed in 0c5fa6b

bram85:
You can squash the commits with the functionality and the documenting comments.

PR #98, opened 2016-01-02
Execute commands on multiple items + repeat command

mruwek:
1. *m* is the default key for toggling 'marked' items.
2. Marking works now across all columns.
3. \<Esc\> will clear the whole selection.
4. Command will be executed on **all** items in **all** columns.
5. *.* repeats last command on currently selected item or marked items.
6. *repeat* remembers silent/verbose output.

I'm using it daily for about 2 weeks and didn't find any major bug. The only minor drawback is that *mark* action uses the same `markup` function so there is no visual difference between *focused* and *marked*. I didn't decide just yet how to deal with it. I see 2 solutions:

1. Use other colors to show *marked* status.
2. Extend current `markup` function to visualize focus on *marked* items differently than on unmarked ones.

### DIFF
--- a/topydo/ui/Main.py
+++ b/topydo/ui/Main.py
@@ -169,14 +169,19 @@ class UIApplication(CLIApplicationBase):
     def _output(self, p_text):
         self._print_to_console(p_text + "\n")
 
-    def _execute_handler(self, p_command, p_output=None):
+    def _execute_handler(self, p_command, p_todo_id=None, p_output=None):
         """
         Executes a command, given as a string.
         """
-        if '{}' in p_command:
-            todos = ' '.join(self.marked_todos)
-            p_command = p_command.format(todos)
         p_output = p_output or self._output
+
+        self._last_cmd = (p_command, p_output == self._output)
+
+        if '{}' in p_command:
+            if self._has_marked_todos():
+                p_todo_id = ' '.join(self.marked_todos)
+            p_command = p_command.format(p_todo_id)
+
         p_command = shlex.split(p_command)
         (subcommand, args) = get_subcommand(p_command)
 
@@ -211,29 +216,14 @@ class UIApplication(CLIApplicationBase):
         if dirty or self.marked_todos:
             self._reset_state()
 
-    def _save_cmd(self, p_cmd, p_execute_signal):
-        verbosity = False
-        if p_execute_signal == 'execute_command':
-            verbosity = True
-        self._last_cmd = (p_cmd, verbosity)
-
     def _repeat_last_cmd(self, p_todo_id=None):
         try:
             cmd, verbosity = self._last_cmd
         except TypeError:
             return
 
-        if verbosity:
-            output = self._output
-        else:
-            output = lambda _: None
-
-        if '{}' in cmd:
-            if not p_todo_id:
-                p_todo_id = ' '.join(self.marked_todos)
-            cmd = cmd.format(p_todo_id)
-
-        self._execute_handler(cmd, output)
+        self._execute_handler(cmd, p_todo_id,
+                              self._output if verbosity else lambda _: None)
 
     def _reset_state(self):
         self.marked_todos = []
@@ -360,13 +350,13 @@ class UIApplication(CLIApplicationBase):
         When no position is given, it is added to the end, otherwise inserted
         before that position.
         """
+        def execute_silent(p_cmd, p_todo_id=None):
+            self._execute_handler(p_cmd, p_todo_id, lambda _: None)
 
         todolist = TodoListWidget(p_view, p_view.data['title'], self.keymap)
-        no_output = lambda _: None
         urwid.connect_signal(todolist, 'execute_command_silent',
-                             lambda cmd: self._execute_handler(cmd, no_output))
+                             execute_silent)
         urwid.connect_signal(todolist, 'execute_command', self._execute_handler)
-        urwid.connect_signal(todolist, 'save_cmd', self._save_cmd)
         urwid.connect_signal(todolist, 'repeat_cmd', self._repeat_last_cmd)
         urwid.connect_signal(todolist, 'refresh', self.mainloop.screen.clear)
         urwid.connect_signal(todolist, 'add_pending_action', self._set_alarm)
@@ -375,8 +365,6 @@ class UIApplication(CLIApplicationBase):
         urwid.connect_signal(todolist, 'show_keystate', self._print_keystate)
         urwid.connect_signal(todolist, 'toggle_mark',
                              self._process_mark_toggle)
-        urwid.connect_signal(todolist, 'has_marked_todos',
-                             self._has_marked_todos)
 
         options = self.columns.options(
             width_type='given',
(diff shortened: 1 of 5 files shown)


topydo/ui/Main.py
```diff
@@ -95,6 +95,8 @@ def __init__(self):
         self.todofile = TodoFile.TodoFile(config().todotxt())
         self.todolist = TodoList.TodoList(self.todofile.read())
 
+        self.pending_todos = []
```
bram85: `marked_todos` would be a clearer name, IMHO.

topydo/ui/Main.py
```diff
@@ -190,17 +197,47 @@ def _execute_handler(self, p_command, p_output=None):
...
+        except TypeError:
+            return
+
+        if verbosity:
+            output = self._output
+        else:
+            output = lambda _: None
+
+        if '{}' in cmd:
+            if not p_todo_id:
```
bram85: Please add some comments here that you handle multiple selection here. Reading this code was not immediately obvious to me.

topydo/ui/Main.py
```diff
@@ -190,17 +197,47 @@ def _execute_handler(self, p_command, p_output=None):
...
+                p_todo_id = ' '.join(self.pending_todos)
+            cmd = cmd.format(p_todo_id)
+
+        self._execute_handler(cmd, output)
```
bram85:
```python
self._execute_handler(cmd, self.output if verbosity else lambda _: None)
```

and remove the verbosity block above.

topydo/ui/Main.py
```diff
@@ -449,6 +491,20 @@ def _console_width(self):
 
         return sz
 
+    def _is_pending(self):
+        if len(self.pending_todos) > 0:
```
bram85:
```python
return len(self.pending_todos) > 0
```

topydo/ui/Main.py
bram85: Besides, if you make it a `set()`,  you don't need to worry whether a todo item was already present in the pending list (in the code below). As far as I can see, order of the todo items is not important in this context.

topydo/ui/Main.py
```diff
@@ -449,6 +491,20 @@ def _console_width(self):
...
+            return True
+        else:
+            return False
+
+    def _pending_todos_handler(self, p_todo_id):
```
bram85: This function does two things, but that's not apparent from its name (again, the verb thing I mentioned somewhere else earlier). I suspect that this function should be split and do only one thing.

topydo/ui/TodoListWidget.py
```diff
@@ -240,6 +261,8 @@ def resolve_action(self, p_action_str, p_size=None):
                 self._execute_on_selected(cmd, execute_signal)
             else:
                 urwid.emit_signal(self, execute_signal, cmd)
+
+            urwid.emit_signal(self, 'save_cmd', cmd, execute_signal)
         else:
```
bram85: I think this signal is redundant. The main class should save the executed command transparently. Let a column just emit 'repeat last command' and let it not worry whether it was saved.

topydo/ui/TodoListWidget.py
```diff
@@ -328,3 +356,15 @@ def _postpone_selected(self, p_pattern, p_mode):
...
+
+    def _repeat_cmd(self):
+        try:
+            todo = self.listbox.focus.todo
+            todo_id = str(self.view.todolist.number(todo))
+        except AttributeError:
+            todo_id = None
+
+        result = urwid.emit_signal(self, 'check_pending_todos')
+        if result:
```
bram85:
```python
if urwid.emit_signal(self, 'has_pending_todos'):
```

Note that I have changed the signal name here a bit, perhaps a better name.

topydo/ui/TodoListWidget.py
```diff
@@ -199,6 +203,17 @@ def mouse_event(self, p_size, p_event, p_button, p_column, p_row, p_focus):
     def selectable(self):
         return True
 
+    def _append_pending_todos(self):
+        try:
+            todo = self.listbox.focus.todo
+            todo_id = str(self.view.todolist.number(todo))
+            result = urwid.emit_signal(self, 'append_pending_todos', todo_id)
+            attr_spec = {None: markup(todo, result)}
```
bram85: Please add methods `mark` / `unmark` to `TodoWidget`, and let the TodoWidget figure out how it should render itself. This is not the responsibility of the TodoListWidget. So the `markup` function stays internal.

bram85:
> The only minor drawback is that mark action uses the same markup function so there is no visual difference between focused and marked. I didn't decide just yet how to deal with it.

Perhaps todo items with a marking are rendered with a surrounding frame with a white background?

topydo/ui/TodoListWidget.py
mruwek: That makes sense, but it would require significant changes. Right now `UIApplication._execute_handler` gets complete todo command (with potential todo id already supplied) or uncomplete (with `{}` placeholder in place of todo id). Depending on that it is decided to use `pending_todos` or not.

If we want to remove the signal above we have to:
1. force `TodoListWidget._execute_on_selected` to **always** send id of focused todo item.
2. Decide upon `pending_todos` vs sent todo id based on value of `pending_todos`
3. Change meaning (and type) of `p_command` argument in `UIApplication._execute_handler` to tuple containing command string (with potential `{}` placeholder) and potential todo id.
4. force all clients of `UIApplication._execute_handler` (for instance command-line widget) to send tuples instead of strings.

Points 3 and 4 can also be resolved by adding supplemental argument for `UIApplication._execute_handler` but that would of course require to adjust client calls.

topydo/ui/TodoListWidget.py
mruwek: Done in 82cc374.

topydo/ui/TodoListWidget.py
mruwek: Ok, I've followed my own second proposition. Results of this approach are visible in 1e1499e.

topydo/ui/Main.py
mruwek: I didn't split it yet, only renamed and documented in 6665efa.

topydo/ui/Main.py
mruwek: Done.

topydo/ui/Main.py
mruwek: Done.

topydo/ui/Main.py
mruwek: I think that after refactoring in 1e1499e those comments are no longer needed.

topydo/ui/Main.py
mruwek: I did the renaming part, but I couldn't see the benefit of transition to `set()`. In current implementation we have to worry whether a todo item is already present in the marked list, as this is the best way to provide a toggle of the mark.

topydo/ui/Main.py
```diff
@@ -191,17 +203,32 @@ def _execute_handler(self, p_command, p_output=None):
             # TODO: show error message
             pass
 
-    def _update_all_columns(self):
+    def _update_all_columns(self, p_reset_keystate=False):
```
bram85: Is the new parameter really required? What could break if we reset it unconditionally?

mruwek: After giving it a thought I removed it in latest rebase.

bram85:
Great work, thanks.

PR #100, opened 2016-01-07
Don't concatenate column layout configs

mruwek:
Use only one layout file. Following precedence is applied:
1. Local (i.e. $PWD/.topydo_columns or
   $PWD/topydo_columns.conf or $PWD/topydo_columns.ini
2. User's home (i.e. $HOME/.topydo_columns or
   $HOME/.config/topydo/columns)
3. System-wide (i.e. /etc/topydo_columns.conf)

### DIFF
--- a/topydo/ui/ColumnLayout.py
+++ b/topydo/ui/ColumnLayout.py
@@ -41,15 +41,18 @@ def columns():
 
     cp = configparser.RawConfigParser(defaults)
     files = [
-        "/etc/topydo_columns.conf",
-        home_config_path('.config/topydo/columns'),
-        home_config_path('.topydo_columns'),
-        ".topydo_columns",
-        "topydo_columns.conf",
         "topydo_columns.ini",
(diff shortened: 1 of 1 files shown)


topydo/ui/ColumnLayout.py
```diff
@@ -41,15 +41,18 @@ def _get_column_dict(p_cp, p_column):
 
     cp = configparser.RawConfigParser(defaults)
     files = [
-        "/etc/topydo_columns.conf",
-        home_config_path('.config/topydo/columns'),
-        home_config_path('.topydo_columns'),
         ".topydo_columns",
```
bram85: Very minor, lines 44-46 should also be reversed, so the various extensions/variants of the names are evaluated in the same order as the main configuration file (Config.py lines 125 - 130).

mruwek: Fixed in fb79242.

PR #108, opened 2016-02-02
Don't crash on shlex.split error

### LINKED ISSUE #107: Apostrophes in prompt mode cause crashes
```
topydo> add read matt's draft
  File "/home/andreas/pool/bin/topydo", line 11, in <module>
    sys.exit(main())
  File "/home/andreas/pool/local/lib/python2.7/site-packages/topydo/cli/UILoader.py", line 45, in main
    PromptApplication().run()
  File "/home/andreas/pool/local/lib/python2.7/site-packages/topydo/cli/Prompt.py", line 103, in run
    user_input = shlex.split(user_input)
  File "/home/andreas/pool/local/lib/python2.7/site-packages/ushlex.py", line 298, in split
    else:       return list(lex)
  File "/home/andreas/pool/local/lib/python2.7/site-packages/ushlex.py", line 283, in next
    token = self.get_token()
  File "/home/andreas/pool/local/lib/python2.7/site-packages/ushlex.py", line 99, in get_token
    raw = self.read_token()
  File "/home/andreas/pool/local/lib/python2.7/site-packages/ushlex.py", line 185, in read_token
    raise ValueError, "No closing quotation"
ValueError: No closing quotation
```

mruwek:
Show some meaningful error message instead.

Fixes #107

### DIFF
--- a/topydo/Commands.py
+++ b/topydo/Commands.py
@@ -21,7 +21,7 @@ instance based on an argument list.
 
 import sys
 
-from topydo.lib.Config import config
+from topydo.lib.Config import config, ConfigError
 
 _SUBCOMMAND_MAP = {
     'add': 'AddCommand',
@@ -90,7 +90,11 @@ def get_subcommand(p_args):
         If alias resolves to non-existent command, main help message is
         returned.
         """
-        real_subcommand, alias_args = alias_map[p_alias]
+        try:
+            real_subcommand, alias_args = alias_map[p_alias]
+        except ValueError as ve:
+            raise ConfigError(alias_map[p_alias]) from ve
+
         try:
             result = import_subcommand(real_subcommand)
             args = join_args(p_args, alias_args)
--- a/test/data/aliases.conf
+++ b/test/data/aliases.conf
@@ -4,3 +4,4 @@ baz = FooBar
 format = ls -F "|I| x c d {(}p{)} s k" -n 25
 smile = ls ☻
 star = tag {} star 1
+quot = lol'd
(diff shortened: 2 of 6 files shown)


bram85:
I think a similar fix is also needed in `Config.aliases()`

mruwek:
> I think a similar fix is also needed in Config.aliases()

Ok, I'll look into it and amend the commit above.

mruwek:
> Would it be possible to return None instead of 'error'

It would cause another problem, as we use `None` whenever we want to show `usage()`.

> What if I'd define an alias called error?

Valid concern. Maybe it should return `False` or `0` (int type)?

bram85:
> Valid concern. Maybe it should return `False` or `0` (int type)?

No, raising an exception would be more appropriate, which can also replace returning `None`.

topydo/Commands.py
```diff
@@ -90,7 +90,11 @@ def resolve_alias(p_alias, p_args):
         If alias resolves to non-existent command, main help message is
         returned.
         """
-        real_subcommand, alias_args = alias_map[p_alias]
+        try:
+            real_subcommand, alias_args = alias_map[p_alias]
+        except ValueError:
+            raise ConfigError(alias_map[p_alias])
```
bram85: I think the exception should be linked with the `as` and `from` syntax.

bram85:
Change looks good bar a minor remark. We must not forget to catch `ConfigError` in the column-ui branch.

topydo/Commands.py
mruwek: Done in b2f34ad.

mruwek:
> We must not forget to catch ConfigError in the column-ui branch.

I would suggest: `topydo add (B) Catch unclosed quote ConfigError in column-ui after:nwh`

PR #109, opened 2016-02-04
Return color config values as integers and update tests for colors

mruwek:
Tests are runnable again.

So far I see 2 problems (thanks to updated tests):

- [ ] numerical values above 255 are processed as normal, instead of returning default or `NEUTRAL_COLOR` resulting in something like `\x1b[4;38;5;777m` for config value `777`.
- [ ] We need some config value to turn 256 color palette on/off as we previously used 256 by default for numerical values and 16 color palette for text values. Right now there is no method (as far as I can see) to get 256 color-palette for ls output.

### DIFF
--- a/test/test_colors.py
+++ b/test/test_colors.py
@@ -19,152 +19,194 @@
 import unittest
 
 from test.topydo_testcase import TopydoTest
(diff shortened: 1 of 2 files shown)


bram85:
Many thanks for those fixes.

PR #110, opened 2016-02-05
Recover previous ansi-color functionality

mruwek:
Seems like everything works now, but 2 nasty hacks are involved:

1. [Colors.py@199](https://github.com/mruwek/topydo/blob/a3651423b15c3e369e872334a345634e5fd828ea/topydo/lib/Colors.py#L199)
2. [Colors.py@61](https://github.com/mruwek/topydo/blob/a3651423b15c3e369e872334a345634e5fd828ea/topydo/lib/Colors.py#L61)

Ideas for cleaner solutions are welcome :)

### DIFF
--- a/topydo/lib/Colors.py
+++ b/topydo/lib/Colors.py
@@ -58,7 +58,13 @@ def get_color(p_type, p_todo, p_256color=False):
         try:
             return color_names_dict[p_input]
         except KeyError:
-            return p_input
+            try:
+                return int(p_input)
+            except ValueError:
+                if p_input:
+                    return -1
+                else:
+                    return p_input
 
     def priority_color():
         priority_colors = config().priority_colors()
@@ -157,7 +163,7 @@ def get_color(p_type, p_todo, p_256color=False):
 
     return normalize_color(result)
 
-def get_ansi_color(p_type, p_todo, p_256color=False, p_background=None, p_decoration='normal'):
+def get_ansi_color(p_type, p_todo, p_background=None, p_decoration='normal'):
     """
     Returns ansi code for color based on xterm color id (0-255) and
     decoration, where decoration can be one of: normal, bold, faint,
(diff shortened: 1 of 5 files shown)


topydo/lib/Colors.py
```diff
@@ -58,7 +58,13 @@ def normalize_color(p_input):
         try:
             return color_names_dict[p_input]
         except KeyError:
-            return p_input
+            try:
```
bram85: With the knowledge I gained in the master branch to optimize topydo, I think we should get rid of the `KeyError` clause, since the exception is likely to be raised when using 256 colors and hurting performance.

```python
try:
  return color_names_dict[p_input] if p_input in color_names_dict else int(p_input)
except ValueError:
  return -1
```

I don't think returning the value as-is is needed in the `ValueError` clause. It should always return an integer (xterm range and -1). The `else` would only be triggered for input values `None` or `-1`.

Unless I overlook something here..

topydo/lib/Colors.py
```diff
@@ -189,13 +193,19 @@ def ansicode(p_int, p_as_background=False):
...
-    bg_color = get_color(p_background, p_todo, p_256color) if p_background else -1
+    fg_color = get_color(p_type, p_todo)
+    bg_color = get_color(p_background, p_todo) if p_background else -1
+
+    try:
+        if 256 > fg_color >= 0:
+            decoration = decoration_dict[p_decoration]
+        else:
+            decoration = '0'
+    except TypeError:
```
bram85: Did you mean `KeyError` here? Why not a `normal` decoration when a bogus decoration is passed to this function?

topydo/lib/Colors.py
mruwek: I believe we need 3 types of values returned here:

1. appropriate color id (valid config or default)
2. `-1` - NEUTRAL_COLOR (invalid config)
3. something that will result in not applying color at all (empty config value)

In my code above this idea is (poorly) implemented with third type of returned value which as string (raw `p_input`).

topydo/lib/Colors.py
mruwek: As I said it's an ugly hack. `TypeError` will be raised when `fg_color` is a string which is the third type of value that demands to **not** use colors at all (for instance with `link_color =` in config).

PR #114, opened 2016-02-14
Use all configurable colors in TodoWidget

mruwek:
Previously only priority color was rendered. Now all contexts, projects,
links and metadata should show theirs colors.

### DIFF
--- a/topydo/ui/Main.py
+++ b/topydo/ui/Main.py
@@ -314,6 +314,7 @@ class UIApplication(CLIApplicationBase):
             current_column.view = view
 
         self._viewwidget_visible = False
+        self._blur_commandline()
 
     def _add_column(self, p_view, p_pos=None):
         """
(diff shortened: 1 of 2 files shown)


topydo/ui/Main.py
```diff
@@ -150,6 +158,39 @@ def hide_viewwidget():
         self.column_mode = _APPEND_COLUMN
         self._set_alarm_for_next_midnight_update()
 
+    def _create_color_palette(self):
+        project_color = _to_urwid_color(config().project_color())
```
mruwek: We have to insert `if config().colors()` here as soon as we get rid of that override above.

bram85:
I think the code is good, and I've been happily using it for over a week now. Please add some comments to the additions in _TodoWidget.py_ (describing your general approach) and then I think it's fine for merge.

PR #115, opened 2016-02-20
Add recurrence and postponing by business days

mruwek:
Inspired and also supported by simpletask[1].

`rec:1b` will recur after 1 business day and `rec:5b` after one full business week. Negative values are supported.

This feature doesn't of course take holidays into account.

[1]: https://github.com/mpcjanssen/simpletask-android/blob/master/src/main/assets/index.en.md


bram85:
Thank you for this feature.

PR #117, opened 2016-03-10
Use business days recurrence in column ui

mruwek:
Using `pp1b` should work now properly.


bram85:
Strange, I remember applying this while I was making merge 8b1ff41ad3. Guess I forgot to `add` it then to the merge commit. Anyway, thanks for applying it.

PR #118, opened 2016-03-12
Provides a smart way to determine the number 'top' lines

### LINKED ISSUE #102: Intelligently Guess the length of prompt
This is in relation to the `top` subcommand. `top` is designed to print out one 'screen-ful' of to-do items. It does this by assuming the prompt takes up 2 lines, and print [number of terminal rows - 2] todo items. However, this falls apart if the prompt is more than 2 lines.

On Windows the `Prompt` environmental variable can be read for `$_` which indicates a newline. So I propose that `top` print one less line for every `$_`. Thoughts?

Is there something equivalent for Linux or MacOS prompts?

MinchinWeb:
Looks at the prompt or PS1 environmental variables to determine the number of
newlines contained in them.

I've tested in on my end in Windows, and it works as expected. I don't have Linux or Mac box, so I haven't been able to test it there.

Because of the way it works (by counting newlines), it won't work exactly right if the prompt goes over the line end without a line break (i.e. for a really long path) or if there are conditional linebreaks in the PS1 variable.

Tests have been added and updated as well.

Fixes #102.

### DIFF
--- a/test/test_list_command.py
+++ b/test/test_list_command.py
@@ -17,6 +17,7 @@
 import codecs
 import re
 import os
+import sys
 import unittest
 from collections import namedtuple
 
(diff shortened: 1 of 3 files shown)


bram85:
Change looks OK. Can you make sure to clean up the history such that `Colors.py` is not in the diff anymore? Then it looks OK to merge.

MinchinWeb:
@bram85 : `Colours.py` "fixed"

bram85:
@MinchinWeb I had another look at your code, and I was wondering why you implemented `_N_lines` in the ListFormat.py module instead of ListCommand.py. The latter is the only client and I don't see much coherence with the formatting functionality (i.e. replacing placeholders). Did you do this for a particular reason? If not, I'd prefer to see this calculation in `ListCommand.py`.

MinchinWeb:
@bram85 : the `_N_lines` function was placed were it was because that seemed to match (my limited understanding of) the organization used in the project. But no worries, I've moved it as you've suggested.

PR #123, opened 2016-04-09
Allow using `tag -a` multiple times

### LINKED ISSUE #121: Using `tag -a` multiple times
If you try and add values for a tag after the second one, `topydo` starts doing weird things. It brings us a prompt asking you which one you want to delete. If you pick a number, it adds the new value. If you pick `all`, it seems to get stuck in a loop.

To reproduce:
```
> topydo add "my item"
|  1| my item
> topydo tag -a k a
|  1| my item k:a
> topydo tag -a k b
|  1| my item k:a k:b
> topydo tag -a k c
 1. a
 2. b
Which value to remove? Enter number or "all": 1
|  1| my item k:a k:b k:c
> topydo tag -a k d
 1. a
 2. b
 3. c
Which value to remove? Enter number or "all": all
[nothing happens till we break out of the program]
```

MinchinWeb:
Fixes #121. Includes regression test.

### DIFF
--- a/test/test_tag_command.py
+++ b/test/test_tag_command.py
@@ -75,6 +75,17 @@ class TagCommandTest(CommandTest):
         self.assertFalse(self.output)
         self.assertEqual(self.errors, "Invalid todo number.\n")
 
+    def test_force_add_tag01(self):
+        '''Tries to different values to a tag for the same name 3 times.'''
+        for letter in ['a', 'b', 'c']:
+            command = TagCommand(['-a', '1', 'k', letter], self.todolist,
+                                 self.out, self.error)
+            command.execute()
+
+            self.assertEqual(self.errors, "")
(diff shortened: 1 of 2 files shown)


topydo/lib/Version.py
```diff
@@ -1,8 +1,8 @@
 """ Version of Topydo. """
 
-VERSION = '0.9'
+VERSION = '0.10-hidden_tag.20160408'
```
bram85: Perhaps you can rewrite the history a bit such that this change does not appear in history (I see you reverted it in a later commit).

bram85:
Thank you for the bugfix, nice catch.

topydo/lib/Version.py
MinchinWeb: I've re-pushed the commits. That should fix it.

PR #125, opened 2016-04-19
Provide adequate "mono" palette for column ui

mruwek:
Focused items are now properly highlighted when colors are set to 0 and marked items are made bold and underlined.

### Before
scrn-before

### After
scrn-after


bram85:
Based on the screenshots, it looks good, but I think we should split the function `_create_color_palette` into `_create_color_palette` and `_create_mono_palette`. The color availability check should be done from the constructor.

mruwek:
@bram85 
I've moved mono palette to it's own method in 40abac4.

PR #126, opened 2016-04-22
Properly print progress color-blocks in ConsoleWidget

mruwek:
2 big changes to the coloring codebase:

1. `topydo.lib.Color` instance is now predetermined to be either foreground or background
2. Progress color is now calculated during `PrettyPrinterColorFilter.filter()` call.

Also fixed one bug that prevented showing colors of link/metadata/project/context when they were located at the start of the output string (e. g. `ls -F %s`)

### DIFF
--- a/topydo/lib/Color.py
+++ b/topydo/lib/Color.py
@@ -45,10 +45,11 @@ class Color:
         'white': 15,
     }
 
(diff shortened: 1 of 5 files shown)


mruwek:
column-ui branch is now merged to master, so closing this one.

PR #128, opened 2016-04-25
Append command: process due dates, start (threshold) dates, and depedancies

MinchinWeb:
I have a habit of appending due dates to existing items on my todo list. However, relative dates weren't being processed. This fixes that.

As well, adding project dependencies via `append` now works as well.

Includes a test case as well, and passes the existing test suite.

### DIFF
--- a/test/test_add_command.py
+++ b/test/test_add_command.py
@@ -140,7 +140,7 @@ class AddCommandTest(CommandTest):
                          self.today + " Bar id:1")
         self.assertEqual(self.errors, "")
 
-    def test_add_de04(self):
+    def test_add_dep04(self):
         """ Test for using an after: tag with non-existing value. """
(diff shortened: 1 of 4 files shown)


MinchinWeb:
c.f. #129

bram85:
I tend to choose the implementation of #129, which introduces a `WriteCommand` that abstracts away the common functionality for `add` and `append`. I like that better than having `TodoList` handle that.

It stays a bit unfortunate that this functionality has been implemented twice, blaming mostly myself for lingering these PRs around for too long.

bram85:
Merged #129

MinchinWeb:
@bram85 : I'm glad you used what you felt was the better code. I had tried to abstract out the common parts between `add` and `append` and couldn't get it to work myself. I'll keep little updates coming as time permits.

PR #129, opened 2016-05-15
Relative dates and dependencies in append cmd

rameshg87:
This commit adds support for relative dates and
dependencies in the append command. Relative dates
in due and start dates will be automatically converted
and dependenices will be recomputed.

When appending due/start dates, if the command already
has due or start dates, it will be replaced.

Change-Id: Ifc2e94e5c7483ea36ccc7b06585e3ca960bc29b5

### DIFF
--- a/test/test_append_command.py
+++ b/test/test_append_command.py
@@ -15,6 +15,7 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import unittest
+from datetime import date
 
 from test.command_testcase import CommandTest
 from topydo.commands.AppendCommand import AppendCommand
@@ -26,6 +27,7 @@ class AppendCommandTest(CommandTest):
         super().setUp()
         self.todolist = TodoList([])
         self.todolist.add("Foo")
+        self.today = date.today().isoformat()
 
     def test_append1(self):
         command = AppendCommand([1, "Bar"], self.todolist, self.out,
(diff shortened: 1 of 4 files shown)


bram85:
Thank you for your contribution. I'm a bit behind with pull requests, but @MinchinWeb implemented the same feature in #128. So now I have to figure out which one to choose :)

I didn't have the chance yet to have a thorough look at the code of both pull requests, but maybe these two can be consolidated to one.

rameshg87:
Hello,

First, I am sorry for not checking existing pull request to see if that feature has been proposed already or not. But to make impartial decision now between the two pull requests for the same thing, I would rather let you decide which pull request to pick.

The CI failure above seems to be because of prompt-toolkit module:
E: 25, 0: Unable to import 'prompt_toolkit.shortcuts' (import-error)

I will try to look into fixing this soon.

topydo/lib/Command.py
```diff
@@ -92,3 +96,64 @@ def usage(self):
     def help(self):
         """ Returns the help text for this command. """
         raise NotImplementedError
+
+
+class WriteCommand(Command):
```
bram85: Please move this class to a separate file (along with the introduced imports above).

bram85:
A great first contribution, thank you!

I have one comment, which shouldn't be too hard to fix, that is to move the `WriteCommand` to a separate file. That way the `import` statements in the generic `Command` file will be reduced to a minimum.

The prompt-toolkit issue should be fixed if you rebase to the latest master.

rameshg87:
Thanks for the review and showing interest in the patch. I have addressed your comment, please take a look at it now.

bram85:
Thanks for your patch

PR #130, opened 2016-06-02
Allow `--version` as an alias for `-v`

MinchinWeb:
Add a long version option.

I keep trying to run `topydo --version`, so I figured I'd make it work!


bram85:
Thanks. Actually I should start using `argparse` to simplify things (such as usage texts), but that's another matter.

PR #131, opened 2016-06-03
Add HiddenTagFilter

MinchinWeb:
Use the `h:1` tag to hide items from being listed

Includes tests (existing tests were adjusted to test for the expected behavior)

### DIFF
--- a/topydo/lib/Filter.py
+++ b/topydo/lib/Filter.py
@@ -160,17 +160,14 @@ class HiddenTagFilter(Filter):
         """
         Constructor.
 
-        A filter which selects a number of Todo instances from a TodoList
-        instance.
-
-        This is handy for constructing a view given a plain list of Todo items.
+        A filter which hides items marked with the approriate tag.
         """
         super().__init__()
         self.todos = p_todos
 
     def match(self, p_todo):
         """
-        Returns Ture when p_todo doesn't have a tag to mark it as hidden.
+        Returns True when p_todo doesn't have a tag to mark it as hidden.
         """
         for my_tag in config().hidden_item_tags():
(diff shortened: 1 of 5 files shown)


topydo/lib/Filter.py
```diff
@@ -154,6 +155,32 @@ def match(self, p_todo):
             return False
 
 
+class HiddenTagFilter(Filter):
+    def __init__(self, p_todos):
+        """
+        Constructor.
+
+        A filter which selects a number of Todo instances from a TodoList
```
bram85: The documentation matches the preceding filter.

MinchinWeb:
@bram85 : Fixes added!

MinchinWeb:
Rebased to v0.10 release

topydo/lib/Filter.py
```diff
@@ -154,6 +155,29 @@ def match(self, p_todo):
...
 
+class HiddenTagFilter(Filter):
+    def __init__(self, p_todos):
+        """
+        Constructor.
+
+        A filter which hides items marked with the approriate tag.
+        """
+        super().__init__()
+        self.todos = p_todos
```
bram85: `self.todos` is not used, so I'd remove it from the constructor.

bram85:
Looks good as far as I can see, bar a minor remark on an unused variable.

Personally I use the `h` tag to hide an item just for one day, for instance I write `h:2016-08-24` in a todo item and have a column defined in column mode with filter `+Project -h:today`. Could be a future extension.

MinchinWeb:
@bram85 : I've removed the unused variable. I think we're good to go then!

bram85:
Thanks :)

PR #133, opened 2016-07-11
Fix saving backups when archiving is turned off

### LINKED ISSUE #132: Crash when using `do` with `-a` option
Running `topydo -a do <something>` yields the following result (on topydo installed from pip):

      File "/usr/local/bin/topydo", line 11, in <module>
        sys.exit(main())
      File "/usr/local/lib/python3.5/site-packages/topydo/cli/UILoader.py", line 49, in main
        CLIApplication().run()
      File "/usr/local/lib/python3.5/site-packages/topydo/cli/CLI.py", line 61, in run
        self._post_execute()
      File "/usr/local/lib/python3.5/site-packages/topydo/cli/CLIApplicationBase.py", line 233, in _post_execute
        self.backup.save(self.todolist)
      File "/usr/local/lib/python3.5/site-packages/topydo/lib/ChangeSet.py", line 102, in save
        list_archive = (self.archive.print_todos()+'\n').splitlines(True)
    AttributeError: 'NoneType' object has no attribute 'print_todos'

I get the same error if I set `archive_filename` to be empty in the configuration file.

mruwek:
Use empty list if user left `archive_file` option empty.

This fixes #132

### DIFF
--- a/topydo/lib/ChangeSet.py
+++ b/topydo/lib/ChangeSet.py
@@ -99,7 +99,10 @@ class ChangeSet(object):
 
         current_hash = hash_todolist(p_todolist)
         list_todo = (self.todolist.print_todos()+'\n').splitlines(True)
-        list_archive = (self.archive.print_todos()+'\n').splitlines(True)
+        try:
+            list_archive = (self.archive.print_todos()+'\n').splitlines(True)
+        except AttributeError:
+            list_archive = []
 
         self.backup_dict[self.timestamp] = (list_todo, list_archive,  self.call)
 
--- a/test/test_revert_command.py
+++ b/test/test_revert_command.py
@@ -284,6 +284,20 @@ class RevertCommandTest(CommandTest):
         self.assertEqual(result, [])
         self.assertEqual(self.errors, "")
 
+    def test_revert07(self):
+        """ Test backup when no archive file is set """
+        backup = ChangeSet(self.todolist, None, ['add One'])
+        backup.timestamp = '1'
+        command1 = AddCommand(["One"], self.todolist, self.out, self.error, None)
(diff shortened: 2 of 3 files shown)


mruwek:
I will add tests and docstrings after positive review of the solution.

topydo/lib/ChangeSet.py
```diff
@@ -99,7 +99,10 @@ def save(self, p_todolist):
 
         current_hash = hash_todolist(p_todolist)
         list_todo = (self.todolist.print_todos()+'\n').splitlines(True)
-        list_archive = (self.archive.print_todos()+'\n').splitlines(True)
+        try:
+            list_archive = (self.archive.print_todos()+'\n').splitlines(True)
+        except AttributeError:
+            list_archive = []
```
bram85: By looking at the code, this would erase the archive file:

```sh
topydo -a do 1
topydo revert
```

If so, my suggestion is to use a copy of the archive content from the previous backup, or the content of the current done.txt if copy is also missing in the changeset.

mruwek:
@bram85 
> By looking at the code, this would erase the archive file:

It wouldn't. `AttributeError` exception is triggered only if `list_archive` is of `NoneType` and that happens only when `config().archive()` returns `False` (cfr [line 286](https://github.com/bram85/topydo/pull/133/files#diff-5781c20f72d8d8a61ce340d7525aef7bR286) in `CLIApplicationBase.py` . I thought that if user sets `archive_file` blank we shouldn't try to be smarter than the user and obey his will to stop saving archive content.

bram85:
In that case, your change looks OK to me.

bram85:
Thanks for fixing this! :)

Issue #134, opened 2016-07-23
Background colors / progress colors don't work in column console 


bram85:
Color blocks, enabled with `%z` in list formatting, are not shown properly in the console of the column UI. They show up as raw ANSI codes instead.

PR #135, opened 2016-07-24
Colorblock in console

mruwek:
This fixes (in a bit hacky way)  #134 

This PR is simply rebase of #126 against master branch. I'm pushing it here to provide a temporary solution and some "food for thougt" about the issue. It probably won't be merged in it's current shape so use at your own risk :)


davesteele:
I personally don't do much with column mode. Before diving into this, please let me know if you are interested in pursuing.

PR #138, opened 2016-08-20
Properly unmark items when resetting state

mruwek:
After introducing cache for TodoWidget objects, widgets highlighted with
'mark' action were not unmarked after 'reset' action and they reappeared
highlighted ("marked") after reverting `do` or `delete` commands.


bram85:
Thanks for the fix.

PR #144, opened 2016-10-23
Fix importance being wrong when due date is a distant monday.

aetherknight:
I discovered a problem today with how importance is calculated (eg, for sorting). Under the following conditions, a task's importance is one higher than it should be:

* I have mostly the default settings (`ignore_weekends` is 1)
* Today is 2016-10-22 (a saturday)
* I have a task due on 2016-11-28 (a monday)

The task shows up with other tasks with an importance of 3 instead of an importance of 2 (based on debugging).

Anyways, this PR tries to address the bug by addressing what looks like an oversight when determining whether to add 1 to the importance if the task is due on Monday.


bram85:
Nice catch, thank you for fixing this :)

Issue #147, opened 2016-11-02
column view alternative configuration file


p-phillips:
I find the columns interface very very useful and intuitive (great piece of work).

I'd like to have an alternative .topydo_columns config file which I could trigger with a command line flag in the same way as the main topydo config file (-x ~/.topydo_columns_dash2)

Had a quick look through the code and couldn't see anything obvious.

Is there code in there already to do this ?

bram85:
No, at the moment you cannot load an alternative file as a commandline parameter, unfortunately.

So with the current version you'd need to work with the various locations where the configuration is looked up (as explained [here](https://topydo.org/#ColumnConfiguration)).

I think, when we implement this, that a column configuration file passed on the commandline should _override_ any other column definition on the standard lookup locations. The current behavior _adds_ columns of one file to the already processed column configuration files.

mruwek:
>  The current behavior adds columns of one file to the already processed column configuration files.

I tend to disagree: https://github.com/bram85/topydo/commit/fb792426de3e35ccf15b4d875584039ff7d74e64

bram85:
Ah, thanks, I thought we left the behavior as you initially implemented it.

bram85:
Implemented in `master`, see commit 0175374.

p-phillips:
Lovely. Just back from leave hence my tardiness in trying this out. I can confirm it works exactly as I'd hoped.

PR #148, opened 2016-11-04
Support user-specified column-layout file

mruwek:
User can now use '-l' option with an argument *after* 'columns' command
and topydo will try to load column layout from that location. If it
won't find specified file it will use standard column-layout file
locations as a fallback.

Example of use:
Will work: `topydo columns -l ~/fooBar_columns`
Won't work: `topydo -l ~/fooBar_columns columns`

Implements #147

### DIFF
--- a/topydo/ui/columns/ColumnLayout.py
+++ b/topydo/ui/columns/ColumnLayout.py
@@ -15,10 +15,12 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import configparser
+from os.path import expanduser
 
 from topydo.lib.Config import home_config_path, config
 
-def columns():
+
+def columns(p_alt_layout_path=None):
     """
     Returns list with complete column configuration dicts.
     """
@@ -49,6 +51,8 @@ def columns():
         "/etc/topydo_columns.conf",
     ]
 
+    if p_alt_layout_path is not None:
+        files.insert(0, expanduser(p_alt_layout_path))
     for filename in files:
         if cp.read(filename):
             break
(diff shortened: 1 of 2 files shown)


topydo/ui/columns/Main.py
```diff
@@ -92,7 +94,23 @@ class UIApplication(CLIApplicationBase):
     def __init__(self):
         super().__init__()
 
-        self._process_flags()
+        f = open('/tmp/dbg-l', 'w+')
+
+        args = self._process_flags()
+        f.write(repr(args))
```
bram85: This temporary file is probably a left-over.

topydo/ui/columns/ColumnLayout.py
```diff
@@ -15,10 +15,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 import configparser
+from os.path import expanduser
 
 from topydo.lib.Config import home_config_path, config
 
-def columns():
+def columns(p_alt_layout_path):
```
bram85: Add a default argument `None` here.

topydo/ui/columns/Main.py
mruwek: Yep. Removed it in last sync.

topydo/ui/columns/ColumnLayout.py
mruwek: Done.

bram85:
Thank you for this feature!

p-phillips:
I commented in another thread, but this is terrific.  Thanks again.

Issue #140, opened 2016-09-29
Customizing colors in columns UI


colinsullivan:
Hello!

Thank you for all of your work!

Is it possible to configure the colors used in the columns UI?

Specifically the highlight color and text changing to black during highlight?

<img width="297" alt="screen shot 2016-09-29 at 9 21 02 am" src="https://cloud.githubusercontent.com/assets/94535/18962530/19692730-8626-11e6-861d-0ca9633afe54.png">

<img width="285" alt="screen shot 2016-09-29 at 9 22 47 am" src="https://cloud.githubusercontent.com/assets/94535/18962584/50a8caf2-8626-11e6-963c-0de34a513fed.png">

bram85:
No, these colors are hardcoded at the moment. But after having a brief look
at the code in shouldn't be too hard to make it configurable.

colinsullivan:
How does this look to you?

I just did the "focus background" colors.

https://github.com/colinsullivan/topydo/commit/a38c763d52c63913f168bcadbb9ea4d41ab5c6c3

bram85:
Can you demo it with a screenshot, please?
Code-wise the patch looks OK, if you make a PR that makes it easier to review, test and merge it.

colinsullivan:
My `.topydo`:

```
...
[columns]
column_width = 79

[colorscheme]
context_color = 6
context_focus_background_color = 8
default_focus_background_color = 8
project_focus_background_color = 8
link_focus_background_color = 8
metadata_focus_background_color = 8
marked_background_color = 16

```

Here is a screenshot:

![screen shot 2016-11-21 at 10 31 10 am](https://cloud.githubusercontent.com/assets/94535/20495848/a7cffb68-afd7-11e6-969b-2371b8659512.png)

Current issue:

Behavior is not correct when I drop one of these new values from the config.  I tried putting the defaults in here: https://github.com/colinsullivan/topydo/commit/ef851a7b83a739e9f8f534ea2e56011eb0bcac20

But this doesn't quite do the trick.

I'm happy to create a pull request with what is there if it would help.

colinsullivan:
On a related note and more of a design choice: I'm not sure anyone would ever want to control these "focus background color" values independently:

```
context_focus_background_color
default_focus_background_color
project_focus_background_color
link_focus_background_color
metadata_focus_background_color
```

Controlling them independently creates square blocks around the different components of the todo.  Perhaps just one `focus_background_color` is more appropriate?

[...]

PR #151, opened 2016-11-21
configurable focus and marked background colors

colinsullivan:
This patch allows the background colors of focused and marked todos to be configured using new configuration options.

With these values:

```
[colorscheme]
focus_background_color = 8
marked_background_color = 16
```

My terminal displays this:

screen shot 2016-11-21 at 11 45 54 am

Defaults are working.

Addresses #140

### DIFF
--- a/test/data/ColorsTest5.conf
+++ b/test/data/ColorsTest5.conf
@@ -4,3 +4,5 @@ project_color =
 context_color =
 link_color =
 metadata_color =
+focus_background_color =
+marked_background_color =
--- a/topydo.conf
+++ b/topydo.conf
@@ -50,6 +50,8 @@ append_parent_contexts      = 0
 ; context_color = magenta
 ; metadata_color = green
 ; link_color = light-cyan
+; focus_background_color': gray,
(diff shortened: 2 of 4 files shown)


bram85:
First of all, thanks for your patch!

I see that the test coverage degraded somewhat, you could fix that by adding the new configuration items to [test/data/ColorsTest5.conf](https://github.com/bram85/topydo/blob/master/test/data/ColorsTest5.conf).

Moreover, when all is in place, please squash the commit history to a single commit to keep the log clean.

colinsullivan:
@bram85 I added config items you suggested, seems like a similar result?

bram85:
I noticed, I'm fixing it up here locally, I do expect to commit and merge by tomorrow. There's still some tests to be added to test/test_{config,colors}.py to cover all code.

bram85:
I added the tests and merged with a [new commit](https://github.com/bram85/topydo/commit/162cc7769505ea455d8f4dac426afeaaebf71713) (also squashed the separate commits).

PR #158, opened 2017-01-16
Allow actions on multiple ids without MultiCommand

mruwek:
This feature means that commands like `tag`, `append` or `dep` can work now with multiple (marked) todo items in column UI.

### DIFF
--- a/topydo/lib/TodoList.py
+++ b/topydo/lib/TodoList.py
@@ -289,4 +289,3 @@ class TodoList(TodoListBase):
         self._depgraph.transitively_reduce()
         clean_parent_relations()
         clean_orphan_relations()
-
--- a/topydo/ui/columns/Main.py
+++ b/topydo/ui/columns/Main.py
@@ -324,6 +324,8 @@ class UIApplication(CLIApplicationBase):
         try:
             if transaction.execute():
                 self._post_execute()
+            else:
+                self._rollback()
         except TypeError:
             # TODO: show error message
             pass
@@ -340,6 +342,12 @@ class UIApplication(CLIApplicationBase):
 
         if dirty or self.marked_todos:
             self._reset_state()
+
+    def _rollback(self):
+        try:
+            self.backup.apply(self.todolist, p_archive=None)
+        except AttributeError:
+            pass
 
     def _repeat_last_cmd(self, p_todo_id=None):
         try:
--- a/topydo/ui/columns/Main.py
+++ b/topydo/ui/columns/Main.py
@@ -38,6 +38,7 @@ from topydo.ui.columns.ConsoleWidget import ConsoleWidget
 from topydo.ui.columns.KeystateWidget import KeystateWidget
 from topydo.ui.columns.TodoWidget import TodoWidget
 from topydo.ui.columns.TodoListWidget import TodoListWidget
+from topydo.ui.columns.Transaction import Transaction
 from topydo.ui.columns.Utils import PaletteItem, to_urwid_color
 from topydo.ui.columns.ViewWidget import ViewWidget
 from topydo.ui.columns.ColumnLayout import columns
@@ -269,11 +270,6 @@ class UIApplication(CLIApplicationBase):
 
         self._last_cmd = (p_command, p_output == self._output)
 
-        if '{}' in p_command:
-            if self._has_marked_todos():
-                p_todo_id = ' '.join(self.marked_todos)
-            p_command = p_command.format(p_todo_id)
-
         try:
             p_command = shlex.split(p_command)
         except ValueError as verr:
@@ -281,24 +277,30 @@ class UIApplication(CLIApplicationBase):
             return
 
         try:
-            (subcommand, args) = get_subcommand(p_command)
+            subcommand, args = get_subcommand(p_command)
         except ConfigError as cerr:
             self._print_to_console(
                 'Error: {}. Check your aliases configuration.'.format(cerr))
             return
 
-        self._backup(subcommand, args)
+        env_args = (self.todolist, p_output, self._output, self._input)
+        ids = None
+
+        if '{}' in args:
+            ids = self.marked_todos if self._has_marked_todos() else [p_todo_id]
+
+        transaction = Transaction(subcommand, env_args, ids)
+        transaction.prepare(args)
 
         try:
-            command = subcommand(
-                args,
-                self.todolist,
-                p_output,
-                self._output,
-                self._input,
-            )
+            self._backup(subcommand, args)
+        except TypeError:
+            self._print_to_console('Error: no todo item was marked for this'
+                                   ' action.')
+            return
 
-            if command.execute() != False:
+        try:
+            if transaction.execute():
                 self._post_execute()
 
         except TypeError:
--- a/topydo/ui/columns/Main.py
+++ b/topydo/ui/columns/Main.py
@@ -301,8 +301,6 @@ class UIApplication(CLIApplicationBase):
                 'Error: {}. Check your aliases configuration.'.format(cerr))
             return
 
-        self._backup(subcommand, args)
-
         env_args = (self.todolist, p_output, self._output, self._input)
         ids = None
 
@@ -320,6 +318,8 @@ class UIApplication(CLIApplicationBase):
 
         transaction = Transaction(subcommand, env_args, ids)
         transaction.prepare(args)
+        label = transaction.label
+        self._backup(subcommand, p_label=label)
 
         try:
             if transaction.execute():
--- a/topydo/ui/columns/Main.py
+++ b/topydo/ui/columns/Main.py
@@ -119,7 +119,7 @@ class UIApplication(CLIApplicationBase):
         self.todofile = TodoFileWatched(config().todotxt(), callback)
         self.todolist = TodoList.TodoList(self.todofile.read())
 
-        self.marked_todos = []
+        self.marked_todos = set()
 
         self.columns = urwid.Columns([], dividechars=0,
             min_width=config().column_width())
@@ -261,6 +261,24 @@ class UIApplication(CLIApplicationBase):
     def _output(self, p_text):
         self._print_to_console(p_text)
 
+    def _check_id_validity(self, p_ids):
+        """
+        Checks if there are any invalid todo IDs in p_ids list.
+
+        Returns proper error message if any ID is invalid and None otherwise.
+        """
+        errors = []
+        valid_ids = self.todolist.ids()
+
+        if len(p_ids) == 0:
+            errors.append('No todo item was selected')
+        else:
+            errors = ["Invalid todo ID: {}".format(todo_id)
+                      for todo_id in p_ids - valid_ids]
+
+        errors = '\n'.join(errors) if errors else None
+        return errors
+
     def _execute_handler(self, p_command, p_todo_id=None, p_output=None):
         """
         Executes a command, given as a string.
@@ -283,26 +301,29 @@ class UIApplication(CLIApplicationBase):
                 'Error: {}. Check your aliases configuration.'.format(cerr))
             return
 
+        self._backup(subcommand, args)
+
         env_args = (self.todolist, p_output, self._output, self._input)
         ids = None
 
         if '{}' in args:
-            ids = self.marked_todos if self._has_marked_todos() else [p_todo_id]
+            if self._has_marked_todos():
+                ids = self.marked_todos
+            else:
+                ids = {p_todo_id} if p_todo_id else set()
+
+            invalid_ids = self._check_id_validity(ids)
+
+            if invalid_ids:
+                self._print_to_console('Error: ' + invalid_ids)
+                return
 
         transaction = Transaction(subcommand, env_args, ids)
         transaction.prepare(args)
 
         try:
-            self._backup(subcommand, args)
-        except TypeError:
-            self._print_to_console('Error: no todo item was marked for this'
-                                   ' action.')
-            return
-
-        try:
             if transaction.execute():
                 self._post_execute()
-
         except TypeError:
             # TODO: show error message
             pass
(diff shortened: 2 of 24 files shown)


topydo/ui/columns/Main.py
```diff
@@ -43,6 +44,60 @@
...
+        todo items contained in _todo_ids attribute. Also sets _subcommand
+        attribute from p_subcommand.
+        """
+        self._subcommand = p_subcommand
+
+        if self._todo_ids:
+            id_position = p_args.index('{}')
+
+            if issubclass(self._subcommand, MultiCommand):
+                p_args[id_position:id_position] = self._todo_ids
```
bram85: You could also write:

```python
p_args[id_position:id_position+1] = self._todo_ids
# and no need for p_args.remove()
```

topydo/ui/columns/Main.py
```diff
@@ -43,6 +44,60 @@
...
+    """
+    This class implements basic handling of executing any subcommand on multiple
+    todo items.
+    """
+    def __init__(self, p_todo_ids=None):
+        self._subcommand = None
+        self._todo_ids = p_todo_ids
+        self._operations = []
+
+    def prepare(self, p_subcommand, p_args):
```
bram85: This parameter `p_subcommand` should not be here I think. A transaction is a sequence of the _same_ commands, I don't think it's necessary to pass it as a parameter of `prepare()`. Since the command to-be executed is a property of the transaction, I'd expect you'd pass it on as part of the constructor.

topydo/ui/columns/Main.py
```diff
@@ -43,6 +44,60 @@
...
+            else:
+                for todo_id in self._todo_ids:
+                    operation_args = p_args[:]
+                    operation_args.insert(id_position, todo_id)
+                    operation_args.remove('{}')
+                    self._operations.append(operation_args)
+        else:
+            self._operations.append(p_args)
+
+    def execute(self, p_todolist, p_out, p_err, p_input):
```
bram85: We could reduce it to a single parameter, making it a bit more maintenance-friendly. In `Main.py`, when creating an `Transaction`, we define the environment first with a lambda:

```python
command_placeholder = lambda op: subcommand(op, p_output, self._output, self._input)
```

All that needs to be filled are the arguments per operation.

We could pass this lambda to the constructor of `Transaction`. This still allows you to check the type of the command for `MultiCommand` in `prepare()`, and easily execute the operation in `execute()` with (`self.command(operation)`).

That would also address the issue above that the subcommand should not be passed for every `prepare()`.

topydo/ui/columns/Main.py
```diff
@@ -43,6 +44,60 @@
 from topydo.ui.columns.ColumnLayout import columns
 
 
+class Transaction(object):
```
bram85: I'd suggest to move the `Transaction` class to a separate file.

mruwek: Done.

topydo/ui/columns/Main.py
mruwek: Done.

topydo/ui/columns/Main.py
mruwek: Oh, boy. That was just too easy. Done.

topydo/ui/columns/Main.py
mruwek: Done. However Consider this approach to the problem you mentioned:

```python
class Transaction(object):
    def __init__(self, p_subcommand=None, p_env_args=None p_todo_ids=None):
        self._multi = issubclass(p_subcommand, MultiCommand)
        self._subcommand = lambda op: p_subcommand(op, *env_args)
        self._todo_ids = p_todo_ids
        self._operations = []
    
    def prepare(self, p_args):
        if self._todo_ids:
            id_position = p_args.index('{}')
            real_subcommand = self._subcommand([])
            
            if self._multi:
                p_args[id_position:id_position + 1] = self._todo_ids
                self._operations.append(p_args)
            else:
                for todo_id in self._todo_ids:
                    operation_args = p_args[:]
                    operation_args[id_position] = todo_id
                    self._operations.append(operation_args)
        else:
            self._operations.append(p_args)
```

topydo/ui/columns/Transaction.py
```diff
@@ -0,0 +1,59 @@
+# Topydo - A todo.txt client written in Python.
+# Copyright (C) 2015 Bram Schoenmakers <bram@topydo.org>
```
bram85: You may use your own name in the current year :)

topydo/ui/columns/Transaction.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        Prepares list of operations to execute based on p_args, list of
+        todo items contained in _todo_ids attribute and _subcommand
+        attribute.
+        """
+        if self._todo_ids:
+            id_position = p_args.index('{}')
+            real_subcommand = self._subcommand([])
+
+            if isinstance(real_subcommand, MultiCommand):
```
bram85: A guiding comment why we make this `MultiCommand` case, would be helpful. For the `edit` command.

topydo/ui/columns/Transaction.py
```diff
@@ -0,0 +1,59 @@
...
+            self._operations.append(p_args)
+
+    def execute(self):
+        """
+        Executes each operation from _operations attribute.
+        """
+        last_operation = len(self._operations) - 1
+        for i, operation in enumerate(self._operations):
+            command = self._subcommand(operation)
+            if command.execute() != False and i == last_operation:
```
bram85: What if one of the intermediate operations fails (return value `False`)? Also, when the last operation fails, `execute()` will have no return value (should return `False`).

Moreover, in terminology, a transaction is something that either fails and rolls back or completes successfully. This transaction is a best effort: try to succeed as many items as we can.

Perhaps it should return the number of successfully processed todo items, or a list of succeded/failed items, so the user can be notified. At this point undecided.

I do wonder how likely it is that a command fails for some items and succeeds for others. The command is the same everytime, except for the todo ID. And the todo ID is not chosen by the user but passed on by topydo.

mruwek: > What if one of the intermediate operations fails (return value False)? Also, when the last operation fails, execute() will have no return value (should return False).

That's a valid concern.

> Moreover, in terminology, a transaction is something that either fails and rolls back or completes successfully. This transaction is a best effort: try to succeed as many items as we can.

That shouldn't be so hard to implement since in case of any fail we can in my opinion just call `self.backup.apply(self.todolist)` from `UIApplication`.

> I do wonder how likely it is that a command fails for some items and succeeds for others. The command is the same everytime, except for the todo ID. And the todo ID is not chosen by the user but passed on by topydo.

In my opinion it is practically imposible with one topydo instance and without any other todo app connected to the same todo.txt file. With concurrent topydo instances or any other app using same todo file – definetely possible:

1. text ids are on
2. mark at least two items
3. go to the command-line
4. in the other topydo instance delete/change (so the id will disappear or change) one of the marked items
5. go back to first instance where you marked items - you should be still in command-line, and one of the marked items is not there at all or is at least not marked anymore. Type: `append {} Foo`
6. There should be one line with updated todo item and `Invalid todo number given` - as You said: best effort.

It could be fixable by detecting if ids in `self.marked_todos` are accessible after update caused by watchdog or maybe (less subtly) just clear marks and kick the user out of the command-line after any watchdog caused update. Still, I would suggest to implement some kind of `self._rollback()`. Just in case.

topydo/ui/columns/Transaction.py
mruwek: I've changed the year.

topydo/ui/columns/Transaction.py
mruwek: Done.

topydo/ui/columns/Transaction.py
mruwek: The only problem with that theoretical `self._rollback()` is catching failing operations to trigger it. As far as I can understand the code, supplying wrong todo ID (`Invalid todo number given.`) produces the same return value of `Command.execute()` as the proper one. If I am not mistaken we would have to rethink return values of the subcommands first.

topydo/ui/columns/Main.py
```diff
@@ -269,36 +270,39 @@ def _execute_handler(self, p_command, p_todo_id=None, p_output=None):
...
-            command = subcommand(
-                args,
-                self.todolist,
-                p_output,
-                self._output,
-                self._input,
-            )
+            self._backup(subcommand, args)
+        except TypeError:
+            self._print_to_console('Error: no todo item was marked for this'
```
mruwek: This error will be displayed if the user would enter in the commandline something like `do {}` without marking any items before (so the `p_todo_id` is `None` and `self._has_marked_todos()` returns `False`).

topydo/ui/columns/Transaction.py
```diff
@@ -0,0 +1,60 @@
...
+            self._operations.append(p_args)
+
+    def execute(self):
+        """
+        Executes each operation from _operations attribute.
+        """
+        last_operation = len(self._operations) - 1
+        for i, operation in enumerate(self._operations):
+            command = self._cmd(operation)
+            if command.execute() != False and i == last_operation:
```
bram85: As discussed previously, `Transaction.execute()` does not always have a return value.

topydo/lib/TodoList.py
```diff
@@ -290,3 +290,9 @@ def clean_orphan_relations():
         clean_parent_relations()
         clean_orphan_relations()
 
+    def ids(self):
+        if config().identifiers() == 'text':
+            ids = self._id_todo_map.keys()
+        else:
+            ids = [str(i + 1) for i in range(self.count())]
+        return tuple(ids)
```
mruwek: Nice idea. I'll update PR at the evening.

topydo/ui/columns/Main.py
```diff
@@ -283,27 +301,31 @@ def _execute_handler(self, p_command, p_todo_id=None, p_output=None):
...
+        self._backup(subcommand, args)
+
         env_args = (self.todolist, p_output, self._output, self._input)
 
         if '{}' in args:
-            ids = self.marked_todos if self._has_marked_todos() else [p_todo_id]
+            if self._has_marked_todos():
+                ids = self.marked_todos
+            else:
+                ids = set(p_todo_id) if p_todo_id else set()
```
mruwek: This line will convert p_todo_id (which is a string) into a set of single characters. For example `'rlr'` will become `{'r', 'l'}`. I will fix that in next push.

topydo/ui/columns/Transaction.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    This class implements basic handling of executing any subcommand on multiple
+    todo items.
+    """
+    def __init__(self, p_subcommand=None, p_env_args=(), p_todo_ids=None):
+        self._multi = issubclass(p_subcommand, MultiCommand)
+        self._cmd = lambda op: p_subcommand(op, *p_env_args)
+        self._todo_ids = p_todo_ids
+        self._operations = []
+        self._cmd_name = p_subcommand.__module__.lower()[16:-7]
```
bram85: This is a duplicate of [line 252 of CLIApplicationBase.py](https://github.com/mruwek/topydo/blob/c1fe991ac4942b80b238e05eb9264c06ca0f01d0/topydo/ui/CLIApplicationBase.py#L252). Since it's a bit obscure (range 16:-7), it deserves to be refactored to a separate function with a descriptive name (perhaps it should be a method of the `Command` base class).

topydo/ui/columns/Main.py
```diff
@@ -269,38 +288,44 @@ def _execute_handler(self, p_command, p_todo_id=None, p_output=None):
...
+            transaction = Transaction(subcommand, env_args, ids)
+        else:
+            transaction = Transaction(subcommand, env_args)
+
+        transaction.prepare(args)
+        label = transaction.label
+        self._backup(subcommand, p_label=label)
+
+        try:
+            if transaction.execute():
```
bram85: Inside a transaction, where the first command succeeds and the second command fails (its `execute()` method returns `False` for whatever reason), it would leave an inconsistent state. The `TodoList` in memory was modified but `self._post_execute()` would not be called (and therefore the first change is not flushed to disk).

I think a transaction needs to store a copy of the `TodoList` instance before it starts executing, and restore that copy whenever a command fails (so the net effect is that Transaction didn't do anything).

topydo/ui/columns/Main.py
```diff
@@ -269,38 +288,44 @@ def _execute_handler(self, p_command, p_todo_id=None, p_output=None):
...
-            if command.execute() != False:
-                self._post_execute()
+            invalid_ids = self._check_id_validity(ids)
+
+            if invalid_ids:
+                self._print_to_console('Error: ' + invalid_ids)
+                return
 
+            transaction = Transaction(subcommand, env_args, ids)
+        else:
```
bram85: One way to eliminate the `else` clause is to initialize `ids` to `None` before line 306 (before the `if`). And then unconditionally create a `Transaction` as on line 318. A minor issue that improves readability IMHO.

topydo/ui/columns/Main.py
mruwek: There's no need to make another copy of `TodoList`. We can reuse the one stored in `backup` attribute

bram85:
Good point. One copy ought to be enough.

>
>

topydo/ui/columns/Transaction.py
mruwek: > (...) perhaps it should be a method of the `Command` base class

Right now it seems so obvious and proper, that I'm embarassed that I didn't think of that myself. It will also solve the problem with `DeleteCommand` saving in backups as `delete` and therefore not paste-friendly (there is no such command as `delete`).

bram85:
Thanks for all your work. A very useful addition to column mode.

PR #163, opened 2017-01-26
Return True after finishing `revert`

mruwek:
Each and every command shall return something at the end of the `Command.execute()`


mruwek:
@bram85 
As I mentioned on gitter: `add`, `append`, `dep`, `lsprj`, `lscon`, `sort`, `tag` shows similiar problem. Waiting for your opinion on that topic.

mruwek:
I'll leave this open for now as a reminder.. I think that after 0.13 is released we should rethink return value of the `Command.execute()` and refactor all commands afterwards.

davesteele:
Whatever else is going on, the code has diverged significantly.

PR #167, opened 2017-02-06
Prefer filterexpr if there is no title in config

mruwek:
Use 'Yet another column' only if there is no title and filterexpr is also empty. This behavior is in accordance with change introduced in fc154914a24b313e2cecbfd55b7147bcaa8ea37e.


bram85:
Looks good, thanks for this fix.

PR #172, opened 2017-02-22
Add possibility to trigger actions after archiving

### LINKED ISSUE #139: 'became active' items are listed with their old ID
If you have items that depend on another, by default they are not shown, but when the dependency is completed, `topydo` will display a message that the item has now become available. However, the id shown (if you are using line numbers) is likely to change as the completed items are archived.

Since that was a mess to write, here's an example:

```
# add the first item
> topydo add first item +test
|484| 2016-08-22 first item +test

# add the second item
> topydo add second item +test
|485| 2016-08-22 second item +test

# make the second item depend on the first
> topydo dep add 484 before 485

# list items; only the first shows
> topydo test
[...]
484   first item +test                                                  (today)

# do the first item
> topydo do 484
Completed: x 2016-08-22 2016-08-22 first item +test p:1
The following todo item(s) became active:
|485| 2016-08-22 second item +test id:1

# above, the second item was shown as being 485. However,
# by the time we can do anything, the item has actually moved
# to being at 484
> topydo test
[...]
484   second item +test                                                 (today)
```

I would expect that the 'item has become active' text should display the item number where it can be accessed for the next command, so after achieving has been accounted for.

mruwek:
Each command can now execute action defined in its own Command.post_archive() method after archiving action is done. Hook for `do` and `del` is also included in this change.

This should fix #139

### DIFF
--- a/topydo/lib/DCommand.py
+++ b/topydo/lib/DCommand.py
@@ -35,9 +35,8 @@ class DCommand(MultiCommand):
             p_args, p_todolist, p_out, p_err, p_prompt)
 
         self.force = False
+        self._delta = []
 
-        # to determine newly activated todos
-        self.length = len(self.todolist.todos())
 
     def get_flags(self):
         return ("f", ["force"])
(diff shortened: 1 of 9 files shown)


topydo/lib/DCommand.py
```diff
@@ -125,4 +123,8 @@ def _execute_multi_specific(self):
                 self.error(self.condition_failed_text())
 
         current_active = self._active_todos()
-        self._print_unlocked_todos(old_active, current_active)
+        self._delta = [todo for todo in current_active
+                       if todo not in old_active]
+
+    def post_archive(self):
```
bram85: Minor: a method name with a verb is more readable: e.g. `execute_post_archive_actions`. Same comment for the new `Transaction` method below.

Issue #149, opened 2016-11-15
Recurring tasks with start dates only


Luxoboy:
Hey there,

Well, first thanks for writing this CLI, greatly appreciated.

I have a feature request: could recurrence be applied to tasks which only have a start date?

For instance, I have a task with a start date in the future. I don't want it to appear or to bother me, hence the start date. When the start date passes on, the task will appear when I list my tasks.
And when I complete the task, I'd like a new one to be created with a start date only (with new_start = completion_date + recurrence_offset).
I hope this makes sense.

Thank you.

bram85:
Actually, the behavior is not built-in. The recurrence code will always set a due date, regardless whether the original has one or not. The start date will only be set when the original has a start date, otherwise its left out.

I started something like this about half a year ago, but abandoned the idea because of corner cases that were hard to come by (I can't remember the details). It had something to do with determining the length of a todo (typically the difference in days between start/creation and the due date).

Moreover, at this point I'm reluctant to add this because it changes the current behavior. Now, items without a due date will get one on recurrence. With this feature this is no longer the case.

Luxoboy:
I understand the change of behaviour is a problem.

What's the reasoning behind the way the start date is set when creating a new recurring task?

Handling the initialization of the start date the same way as the due date would allow me to obtain the behaviour described in the first message.

Thanks!

amariusz:
It's an old issue but I think @Luxoboy request is still valid.
I'd like to let you know I've mentioned this ToPyDo behaviour in other SimpleTask's issue - https://github.com/mpcjanssen/simpletask-android/issues/312#issuecomment-767365840

PR #190, opened 2017-07-22
Generalize postpone to manipulate any tag

wzel:
In my `todo.txt`, I have many records with a `t:` but no `due:` tag. It would be nice to have a way of postponing start dates without introducing a `due:` tag as a byproduct (which is the current behavior of `postpone -s`). Changing the behavior of `postpone -s` would break backward compatibility, so maybe `postpone -s -s`?

Unfortunately, implementing this occurred to be not so trivial, since in the current code the postponing operation derives the new `t:` from the new `due:`, rather than independently. Moreover, missing `t:` is handled differently than missing `due:`.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -10956,12 +10956,13 @@ Business days only take Monday - Friday into account.</pre>
 <pre>Postpone the todo item(s) with the given number(s) and the given pattern.
 
 ```
-postpone [-s] &lt;NUMBER&gt; [&lt;NUMBER2&gt; ...] &lt;PATTERN&gt;&quot;
+postpone [-s [-s]] &lt;NUMBER&gt; [&lt;NUMBER2&gt; ...] &lt;PATTERN&gt;&quot;
 postpone [-x] -e &lt;EXPRESSION&gt;
 ```
 
 Postponing is done by adjusting the [[due date(s)|TagDue]] of the todo(s), and if the `-s`
 flag is given, the [[start date|TagStart]] accordingly.
+If the `-s` flag is repeated, only the [[start date|TagStart]] is adjusted.
 
 It is also possible to postpone items with an expression using
 the `-e` flag. Use `-x` to also process todo items that are normally invisible
--- a/docs/index.html
+++ b/docs/index.html
@@ -10956,13 +10956,14 @@ Business days only take Monday - Friday into account.</pre>
 <pre>Postpone the todo item(s) with the given number(s) and the given pattern.
 
 ```
-postpone [-s [-s]] &lt;NUMBER&gt; [&lt;NUMBER2&gt; ...] &lt;PATTERN&gt;&quot;
+postpone [-s] [-t &lt;TAG&gt;] [-T &lt;TAG&gt;] &lt;NUMBER&gt; [&lt;NUMBER2&gt; ...] &lt;PATTERN&gt;&quot;
 postpone [-x] -e &lt;EXPRESSION&gt;
 ```
 
 Postponing is done by adjusting the [[due date(s)|TagDue]] of the todo(s), and if the `-s`
 flag is given, the [[start date|TagStart]] accordingly.
-If the `-s` flag is repeated, only the [[start date|TagStart]] is adjusted.
+If the -t flag is given adjustment is applied to TAG. If the -T flag is
+given adjustment is applied to TAG under the condition that it exists.
 
 It is also possible to postpone items with an expression using
 the `-e` flag. Use `-x` to also process todo items that are normally invisible
--- a/test/test_postpone_command.py
+++ b/test/test_postpone_command.py
@@ -313,7 +313,7 @@ class PostponeCommandTest(CommandTest):
 
         self.assertTrue(self.todolist.dirty)
         self.assertEqual(self.output, "|  1| Foo\n")
-        self.assertEqual(self.errors, "Warning: todo item has no (valid) due date, therefore it was not adjusted.")
+        self.assertEqual(self.errors, "Warning: todo item has no (valid) due date, therefore it was not adjusted.\n")
 
     def test_postpone26(self):
         command = PostponeCommand(["-T", "t", "1", "1w"], self.todolist, self.out,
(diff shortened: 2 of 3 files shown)


wzel:
Please, just do not merge this. This WiP at this stage: wet, dirty code; just to have something concrete to discuss.

bram85:
Thank you for your contribution! I haven't studied your patch in detail (yet), but my first suggestion would be to use a different character rather than `-s -s`. My suggestion would be `-S`.

Apparently start-date-only tasks are a common use case (see also #149), so I think this is a valid addition to topydo. I will try to have a more thorough look later.

wzel:
Hi, I indeed thought about `-S` initially, but then it occurred to me: what if you happen to provide `-s` and `-S` simultaneously in a single `topydo` call? A conflict error? Introducing `-s -s` avoids the issue. (Thinking about it makes me wonder how different authors handle a possible `--verbose --quiet` flag clash, for instance.)

bram85:
Typically, things like these are documented, if one flag overrides another. I think that's the easiest way out.

Another way is to introduce a new flag that takes an argument, for instance `-S due`, `-S t` to indicate which tag(s) to shift, and is repeatable. So `-s` becomes an alias for `postpone -S due -S t`. In your configuration file you could make an alias for your variant:

```ini
[aliases]
ppt = postpone -S to
```

That makes future extension easier to allow for shifting arbitrary tags. In my todo.txt, I maintain a `planned` tag for some items, holding the date when I'm going to work on that item. With the -S flag I could shift this date with `postpone -S planned 1d`. I'm not asking you to implement this extension, but the more flexible `-S` tag would make it easier.

wzel:
I like the generalized `-S` flag idea. Perhaps some better (lowercase) letter should be reserved for this, as we're not talking about [S]tart, anymore.

Regarding making `-s` a special case of `-S`, the line `self.error("Postponing todo item failed: invalid due date.")` could be generalized, too, removing one of the original obstacles. Unfortunately, there is still this problematic `if self.move_start_date and todo.start_date():` block, which results in a different treatment of `due:` and `t:`. Abandoning this would improve orthogonality, but break backward compatibility.

Perhaps an introduction of two new flags instead of one would solve this problem: one flag (let's call it `--Screate`) creates a tag if it doesn't exist in the first place (the current `due:` behavior), while another one (`--Snocreate`) --- doesn't (the `t:` behavior). Then the alias for `postpone -s` would be `postpone --Screate due --Snocreate t`. IMO, not a high price for making the code more elegant.

I can try to implement this.

bram85:
I think it's a good suggestion. I'd choose `-t` and `-T` (for tag) to specify tags to be created / not to be created respectively.

wzel:
Hopefully, next week I will be able to find time to work on that.

wzel:
There are still some minor issues left:

- [ ] The `t:` tag is not translated into `start` in the error message. Perhaps all non-default tags should be explicitly defined in the config together with their full names and the messages should refer to the full names instead of the tag names.
- [ ] `self.todolist.dirty` should probably not be set to `True` when the `-T` flag raises an error and no adjustment takes place (easy to fix).
- [ ] When more than one flag is given the program will behave... I don't know how (I haven't tried), but probably in some undesirable way.
- [ ] The relevant HTML documentation probably some needs improvement.

bram85:
I think you're on the right track, but it needs a little improvement. It was a bit hard to understand and review it.

The main issue I encountered is that either `-t` or `-T` can be given, and not in combination (or pass each flag more than once, as you already mentioned in the checklist). That these flags override each other is unexpected and not desired.

My proposal is that you maintain two lists of tags, those given with `-t` and those given with `-T`, e.g. `self.move_and_create_tags = []` and `self.move_tags = []` respectively. Then iterate over these two lists and move each tag.

> self.todolist.dirty should probably not be set to True when the -T flag raises an error and no adjustment takes place (easy to fix).

That's right.

I think when these issues are addressed, we introduce some valuable functionality to topydo.

wzel:
Thanks for the valuable comments. This weekend I'm away, but I expect to find enough time during the next one to make the improvements (i.e. rebasing, `self.todolist.dirty` issue, additional tests involving multiple tags, lists).

wzel:
It just occurred to me that using sets might be preferable to using lists. With sets, there is no possibility of a duplicated tag -- also among the sets, since one (the one taking precedence; I would nominate the one associated with `-T` for this) can be substracted from the other before processing.

bram85:
Yes, that's a good suggestion, should save some headaches and lines of code.

wzel:
OK. It kind of works, but ten tests fail, so I cannot push this for you to take a look. The failures can be classified into:

1. Since with multiple tags one tag can raise an error, while postponing of another can work, in the new implementation, the processed todo(s) is/are always printed. This breaks tests 09, 17, 21, 39, 40.

2. The `-s` tag is not treated in a special way, anymore, and thus generic error messages show instead of `-s`-specific ones. This breaks 23.

3. Also with `-s`, the new date for the "t" tag is not calculated in the `-s` specific way (i.e. omitting the comparison with today date), anymore. This breaks 05, 08, 16.

4. Errors involving the "t" tag say "t" rather than "start". This breaks 03, 23.

Regarding the last problem, it would be nice to have some method reversing the `config().tag_*()` family to be able to refer to full names of the tags in messages. There is `_get_tag()` in `Config.py`, but I'm not sure if that's its purpose.

Concerning the remaining issues, perhaps it should be considered whether the new behavior isn't actually better and adjust the tests accordingly.

In September, I'm going to be mostly on the road, so I'm afraid I won't be able to come back to this project before October.

wzel:
I fixed the first issue: If `-T` fails the todo is not echoed, replicating the `-s` behavior with respect to the `t` tag (the behavior which I don't like). The test 17 still fails, though, since the error message is displayed twice -- once for each todo. In my opinion, `pattern` rather than `new_due` should be checked for an error. After all, the error message talks about an "invalid pattern". I can circumvent the problem by using a dirty hack of adding `pattern` to today date and making the check before entering the `for todo in self.todos:` loop. But of course it would be much better to check `pattern` directly.

wzel:
I'm back. The code is backward-compatible now, but also wet, plus warning and error messages are inconsistent:

* `-s` flag and no flag use tag names from `topydo/lib/Config.py`,
* `-t` and `-T` flags use tag names as they appear in the `todo.txt` file.

This could be fixed by having a function generalizing `config().tag_due()` and `config().tag_start()` like `config()tag(p_value)`, where `p_value` would be `"due"`, `"start"`, etc.

It would also make sense to enable the user to extend the tag names dictionary beyond what is defined in `topydo/lib/Config.py`. In https://github.com/bram85/topydo/pull/190#issuecomment-317232515, you mention a `planned` tag, so maybe this is possible, already?

All of this makes me wonder why the start tag isn't simply `start` rather than `t`…

davesteele:
In light of the number of commits involved, and the age of the PR, I'm going to close this. I'd be happy to consider a streamlined rework.

PR #200, opened 2017-10-16
Add auto_uuid feature

janLo:
This adds a feature to assign every newly created task a stable uuid to
identify that task later. The uuid is saved as "uuid" tag field.

Signed-off-by: Jan Losinski <losinski@wh2.tu-dresden.de>

### DIFF
--- a/test/data/ConfigTest4.conf
+++ b/test/data/ConfigTest4.conf
@@ -3,6 +3,7 @@ colors                      = junk
 
 [add]
 auto_creation_date          = junk
+auto_uuid                   = junk
 
 [ls]
 indent                      = junk
--- a/test/test_add_command.py
+++ b/test/test_add_command.py
@@ -395,6 +395,29 @@ class AddCommandTest(CommandTest):
         self.assertEqual(self.todolist.todo(1).source(), "New todo")
         self.assertEqual(self.errors, "")
 
+    def test_add_task_with_auto_uuid(self):
+        config(p_overrides={('add', 'auto_uuid'): '1'})
+
+        args = ["New todo"]
+        command = AddCommand.AddCommand(args, self.todolist, self.out,
+                                        self.error)
+        command.execute()
+
+        self.assertRegex(self.todolist.todo(1).source(), "{tod} New todo uuid:[0-9a-f-]{{36}}".format(tod=self.today))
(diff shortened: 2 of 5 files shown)


mruwek:
Can You provide a use-case for this PR  that is not already possible with built-in [text identifiers](https://topydo.org/#Identifiers)?

janLo:
I'm a fan of todo.txt because of its simplicity and of topydo for its features. Now I try to combine this with a pomodoro-like (or general) timetracking.

I store the time-tracking data in a small database and want to assign it todo.txt tasks. For this I need 100% stable identifiers. So the identifier can not depend on the content, as the content of the task is always changeable (correct typos, ...). Also there should be no duplicates in any case. Not within the current todo.txt file, nor with any past "archived" tasks. UUIDs are one of the most simple ways to achieve that without having the need to a persistent storage of "the last" or "all used" ids.

The text identifiers are nice for human usage within the todo.txt file at a given time. But they are not really suitable for this usecase:
* They are not stable in any case (if the content changes the identifier changes or if there is a duplicate and the first instance is deleted, the the identifier of the second instance change) 
* they are not unique over time. For example take recurring tasks. Every task will have the same text identifier, as the old one gets deleted (archived) and a new one gets created with the same content.

So having another mechanism to identify tasks for such purposes is necessary. If enabled, it could even provide a benefit for the text identifiers, if they use the uuid (if available) to generate the id instead of the content, which can be changed over time or be non-unique.

janLo:
Do you have any requests for changes?

bram85:
I'm not convinced about the use case and I think it could also be established with external tools. You could run the following `sed` command to add a `uuid` tag to those lines that haven't one yet.

```bash
sed -i "/uuid:/! { s/^.*$/echo -n '&' uuid:; uuidgen/e }" todo.txt
```

Notice I'm using a GNU sed feature here with the `s///e` construct, in order to guarantee unique UUIDs for each line.

bram85:
By the way, I'm thinking of an pre/post-hook feature to implement such features in a generic way, like in `git`. So we have a `pre-add` and `post-add` scripts that get executed on todo items.

davesteele:
I'm also not convinced about the use case. It takes the product too far away from its roots.

I wouldn't mind seeing code for the hooks, though.

Issue #227, opened 2018-03-14
[feature idea] Changing day-switch time for something else than midnight


Ekleog:
Just an idea: it could be possible to change day-switch time from midnight to eg. 4pm, for late-sleepers. This way, during [midnight, 4pm], the current day would actually be “yesterday”, allowing to have a “due today” task list that is at the right day even during this interval.

What do you think about it?

davesteele:
Covered by #230

PR #229, opened 2018-03-22
Move all time-related calls to a special helper

Ekleog:
This moves (hopefully) all the code that calls date-related functions to a special helper file.

This should be the first move to support moving the midnight barrier, for https://github.com/bram85/topydo/issues/227: now, there is a centralized place to offset the day if required.

These changes don't break any test here, then it's not adding any feature yet, so that was to be expected.

### DIFF
--- a/topydo/commands/AddCommand.py
+++ b/topydo/commands/AddCommand.py
@@ -18,12 +18,12 @@
 
 import codecs
 import re
-from datetime import date
 from os.path import expanduser
 from sys import stdin
 
 from topydo.lib.Config import config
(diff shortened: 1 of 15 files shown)


Ekleog:
... Ok, I don't get what's going wrong with the `isort` on travis-ci: if I run `isort -c -rc .` in <repository>/topydo, I get only “Skipped 10 files” as a result. If I run it in <repository>, I get a lot of errors from <repository>/lib/python3.6, and then “LookupError: unknown encoding: lala”.

Anyway, I can't reproduce locally Travis-CI's error, so if you'd help me bring this PR in a merge-able state… :/

Ekleog:
Superseded by #230.

PR #230, opened 2018-03-25
Time shift

Ekleog:
This PR adds a `time_shift` configuration parameter in `[topydo]`, that allows to shift the time of day switch.

This completes #227.

Before spending time on documentation and tests, could I just have some feedback on the technical part? :) (I also guess some linters won't pass like in #229, but same thing, that's for later when I know I'm not completely wrong even doing this.)

### DIFF
--- a/topydo/commands/AddCommand.py
+++ b/topydo/commands/AddCommand.py
@@ -18,12 +18,12 @@
 
 import codecs
 import re
-from datetime import date
 from os.path import expanduser
 from sys import stdin
 
 from topydo.lib.Config import config
 from topydo.lib.prettyprinters.Numbers import PrettyPrinterNumbers
+from topydo.lib.Time import today
 from topydo.lib.WriteCommand import WriteCommand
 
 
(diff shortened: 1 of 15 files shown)


Ekleog:
I've just rebased and pushed an additional commit that should take care of an issue I was having with midnight-switch-time. And an additional commit that just was a run of `isort -rc .`, hoping this time the test will pass :)

Thanks for your maintaining this project!

Ekleog:
(I can now confirm that the “Fix midnight update time” had the effect of making auto-refresh-at-day-switch work again)

davesteele:
I closed the issue because the topic is covered here.

I haven't merged this because I'm not convinced that there is a significant demand for this. I could be convinced otherwise with some independent calls here.

I respect that you didn't want to commit to the documentation early, but don't minimize that. It can be difficult on this project.

Ekleog:
I totally understand, and thank you for your work in maintaining topydo!

I guess I'll close this PR then, and if other people want to have this feature they can voice it and we can reopen, there's no need to keep this open for nothing :)

Anyway, good luck with the maintenance of topydo!

PR #252, opened 2019-01-07
Add a setup.py test req for freezegun

davesteele:
Tell setup.py that freezegun is required for tests.

BTW, I'm getting ~15 test failures that appear to be related to Debian bug [916702](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=916702), an incompatibility between older freezegun and Python 3.7 (Debian unstable).


davesteele:
rebased and merged

PR #268, opened 2020-01-22
edit: Don't move modified todo items to the end of file

navytux:
Before this patch a `t edit ...` will remove todo items selected by
`...` and then do equivalent of `t add` with input from result of
editing. This works, but has the property that whenever a todo item is
edited, it is always moved till the end of todo.txt file.

This property creates noise and makes synchronization harder to implement for
some of us who uses VCS and different machines to maintain todo.txt .

-> Teach edit to preserve todo items in their places. Adjust existing
tests and add a couple more exercising cases where either new todo
items are added in the editor, or vice versa, removed in the editor.

/cc @mruwek

### DIFF
--- a/test/test_edit_command.py
+++ b/test/test_edit_command.py
@@ -58,7 +58,7 @@ class EditCommandTest(CommandTest):
 
         self.assertEqual(self.errors, "")
         self.assertTrue(self.todolist.dirty)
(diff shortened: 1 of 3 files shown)


navytux:
( travis failures on `test_revert_ls` are completely unrelated as those failures, unfortunately, are also currently present on master )

navytux:
Thanks, @davesteele

Issue #274, opened 2020-04-02
test_revert_command.test_revert_ls() fails on arrow >= 0.15.0


drewrisinger:
I traced the error back to https://github.com/bram85/topydo/blob/b59fcfca5361869a6b78d4c9808c7c6cd0a18b58/topydo/commands/RevertCommand.py#L101.

Per https://arrow.readthedocs.io/en/latest/releases.html#id9, the command ``arrow.get(timestamp_string)`` doesn't work anymore, and preferred syntax is now ``arrow.get(timestamp_string, "X")``. In this test case, ``timestamp_string="5"`` etc. It seems that ``arrow`` doesn't work with timestamps that are relatively low, due to parsing ambiguity I presume.

The fix will probably need two things:
* convert ``arrow.get(str)`` -> ``arrow.get(str, "X")``
* stop using ridiculously low timestamps for the revert tests. Using something arbitrary in the range of years 2000+ would presumably work.

See #271 for same issue.

davesteele:
Tests pass with arrow 0.16.0 and 60b0e1b. Do you feel more is needed?

drewrisinger:
Can confirm that tests pass when v0.13 is patched with https://github.com/topydo/topydo/commit/60b0e1beb05d975a0509d47232348a9a2fba1810. Thanks!

mruwek:
@davesteele 
> Tests pass with arrow 0.16.0 and [60b0e1b](https://github.com/topydo/topydo/commit/60b0e1beb05d975a0509d47232348a9a2fba1810). Do you feel more is needed?

I'm afraid you missed the real bug. Try executing `revert ls` with arrow>=0.15.0 after the commit above. I'm sure topydo will crash.

You "fixed" the test in a way that it doesn't simulate **topydo**. Topydo backup functionality doesn't save changesets with human-readable dates. It still [uses timestamps](https://github.com/topydo/topydo/blob/0ccb913df6c902964cb7dc0d4ab8db3c387d7217/topydo/lib/ChangeSet.py#L49). Right now test is using human-readable date and the topydo itself is using timestamps. The resulting fix doesn't fix anything. It just hides the problem and pretend that everything is ok :)

The one and only bug is this – that topydo uses timestamp as `string` while it should use `float` there. I'll post the working solution in about an hour.

davesteele:
A little bit about myself:

[...]

PR #280, opened 2020-11-17
Fix `revert ls`

mruwek:
As discussed in #274 the test was actually ok. It was rather the `revert ls` subcommand that needed a tiny fix :)


mruwek:
Tests can break before merging #279. After that should be ok.

Issue #223, opened 2018-02-27
columns view is always one line too short


mbroedl:
I cannot scroll down any further than the first line of the last task of a column.
![image](https://user-images.githubusercontent.com/19365355/36746503-62991842-1bea-11e8-9497-94238b03b269.png)
The only time (it seems) when I can actually see that line is when the last item is selected and the view extends (e.g. after committing a change and then the status line on the bottom disappears).
![image](https://user-images.githubusercontent.com/19365355/36746496-5d308a2a-1bea-11e8-9ea9-89bb0feb98f7.png)

topydo 0.13, python 3.6.4

mruwek:
Yep, that's a bug related to our urwid/urwid usage. 

> The only time (it seems) when I can actually see that line is when the last item is selected and the view extends (e.g. after committing a change and then the status line on the bottom disappears).

It also extends when You use "end" key (default is `G`).

RonCam:
Thanks to @mruwek for keeping an eye on the _Issues_ and posting this work-around.  I was also wondering what to do about the missing lines.  It's to be added to my topydo 'cheat-sheet' card that I keep on my desk.

For users who may not scan the _Issues_ for solutions to problems, might also be a good idea to add it to the _Documentation_, if it's not already there.

mruwek:
I have a fix for this issue in my local tree, but I have a question for current maintainers before I'll post a PR. Fix that I made simply changes how we execute `down` action inside `TodoListWidget`. This will show potential inconsistency - it will work with whatever user sets to `down` action inside `.topydo` (like `j` which is default), but it won't change `<Down>` key result. Do we want to resolve this in `Config.py` by adding:

```
'column_keymap': {
           ...
          '<Down>': 'down',
}, 
```

or internally inside `TodoListWidget.py`?

PR #281, opened 2020-11-23
Better focus in column mode

mruwek:
This PR does 3 things:

- [x] Properly detects first task in column when `home` action is executed and stops there. Previously it focused on divider between first task and title widget.
- [x] Show the task at the bottom of the column in full.  (fixes: #223)
- [x] Highlight title widget so the user always knows where to move next, even when currently focused column is empty.


davesteele:
FYI, I plan on releasing before the end of the year.

PR #286, opened 2021-01-30
Respect auto_creation_date when advancing recurring tasks

niconfus:
This PR fixes the problem: when running `topydo do` on a recurring task, the newly created replacement has the creation date even when `auto_creation_date = 0` is configured.


davesteele:
Thanks

PR #300, opened 2022-09-10
Color relative due date based on remaining time

### LINKED ISSUE #299: Color Relative due date based on remaining Time
I want the relative due date (`%D`) to be
* Colored red when I have less than 24 hours.
* Colored yellow when I have less than 72 hours.
* Colored green when I have more than that.

Or something like that. At the very least I want it to be highlighted in _some way_ so that I can easily distinguish it from the text of the todo item in question.

Unlike most people I actually look through the code to see how my desired feature would be implemented, and it's not quite trivial. All the coloring is put into the last filter for a reason, and I'm not sure if having it in the parser will mess up a step inbetween. However it seems like parsing is the second-to-last filter used anyways, so I don't forsee there being much of a problem.

Anyways, I'm not 100% sure, so I wanted to bring it up here to see your thoughts on the matter. BTW thanks for making the best todo application!

BioBox:
I'm quite happy with how this turned out and it passes all the tests, but there needs to be some unit tests for this which unfortunately I don't have time for. I've used this on my own set of todos and it works just fine.

Fixes #299

### DIFF
--- a/topydo/lib/ListFormat.py
+++ b/topydo/lib/ListFormat.py
@@ -17,9 +17,11 @@
 """ Utilities for formatting output with "list_format" option."""
 
 import re
+from datetime import date, timedelta
 
 import arrow
 
+from topydo.lib.Color import Color
 from topydo.lib.Config import config
 from topydo.lib.ProgressColor import progress_color
 from topydo.lib.Utils import escape_ansi, get_terminal_size, humanize_date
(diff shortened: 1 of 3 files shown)


davesteele:
Looks reasonable at first blush. I have limited access for a bit - I'll address it when I'm more online.

davesteele:
Note the test failures.

BioBox:
Yeah I know only 3.10 because of the match statement.

davesteele:
Changes need to be compatible with [supported](https://devguide.python.org/versions/) Python versions. I'll leave the PR open in the interim.

PR #302, opened 2022-09-29
Fix typo on ls format specifiers

BioBox:
The `%T` and `%t` specifiers are for the date when the todo item is _active_, not when it was created.


davesteele:
Thanks

PR #303, opened 2022-10-01
Add ability to remove priorities.

BioBox:
Use the `-d` flag to delete priorities.


davesteele:
Note that this should be documented in the TiddlyWiki as well.

PR #305, opened 2023-01-01
Add priority flag to ls command

BioBox:
Used like `topydo ls -p A,C,D`. Should work with priority expressions as well as raw capital letters.
I'm can't edit the wiki, so that still needs to be added in, but it seems to work well.


davesteele:
Can you add documentation for this to the tiddlywiki?

BioBox:
Do I just edit the raw html?

davesteele:
Open it with a browser, with JavaScript enabled.

BioBox:
Oh! I had no idea!
Yeah I passed on that because I never edit raw html. I've opened the wiki before but never saw the button.

davesteele:
Sounds good. It may be a week before I can give this a good look.

topydo/commands/ListCommand.py
```diff
@@ -101,6 +104,9 @@ def _process_flags(self):
 
                 # when a user requests a specific ID, it should always be shown
                 self.show_all = True
+            elif opt == '-p':
+                self.priorities = [f"(={p})" if len(p) == 1 else p for p
```
davesteele, line 108: Based on the tiddlywiki edit, I would expect this to detect a more complex expression.

test/test_list_command.py
```diff
@@ -479,6 +479,20 @@ def test_list52(self):
...
+        self.assertEqual(self.output, "|4| (C) Drink beer @ home\n")
+        self.assertEqual(self.errors, '')
+
+    def test_list54(self):
+        command = ListCommand(["-p", "D"], self.todolist, self.out, self.error)
+        command.execute()
+
+        self.assertEqual(self.output, "|2| (D) Bar @Context1 +Project2\n")
+        self.assertEqual(self.errors, '')
+
```
davesteele, line 495: Need a test for e.g. (<B).

davesteele (changes requested):
Looks pretty good. Just need to be consistent on expression support.

PR #323, opened 2024-07-13
Fix recurrence of tasks with start date and no due date

fuzzdk:
Before this change topydo would make a due date on a task without a length giving it a length of 0 days:

```
$topydo add test t:yesterday rec:1d
|  1| 2024-07-13 test t:2024-07-12 rec:1d
$topydo do 1
Completed: x 2024-07-13 2024-07-13 test t:2024-07-12 rec:1d
$cat todo.txt 
2024-07-13 test t:2024-07-14 rec:1d due:2024-07-14
```

with this change it no longer will make a due date in this case:
```
$topydo add test t:yesterday rec:1d
|  1| 2024-07-13 test t:2024-07-12 rec:1d
$topydo do 1
Completed: x 2024-07-13 2024-07-13 test t:2024-07-12 rec:1d
$cat todo.txt 
2024-07-13 test t:2024-07-14 rec:1d
```

This make topydo behave the same way as https://github.com/mpcjanssen/simpletask-android

### DIFF
--- a/test/test_recurrence.py
+++ b/test/test_recurrence.py
@@ -135,8 +135,22 @@ class RecurrenceTest(TopydoTest):
 
         new_start = date.today() + timedelta(7)
         new_todo = advance_recurring_todo(self.todo)
+        new_due = date.today() + timedelta(7)
 
         self.assertEqual(new_todo.start_date(), new_start)
+        self.assertEqual(new_todo.due_date(), new_due)
+
+    def test_startdate4(self):
+        """ Start date and no due date. """
+        self.todo.set_tag(config().tag_start(), date.today().isoformat())
+
+        new_start = date.today() + timedelta(7)
+        new_todo = advance_recurring_todo(self.todo)
+
+        self.assertEqual(new_todo.start_date(), new_start)
+        self.assertEqual(new_todo.due_date(), None)
+
+
 
     def test_strict_recurrence1(self):
         """
--- a/test/test_recurrence.py
+++ b/test/test_recurrence.py
@@ -178,6 +178,19 @@ class RecurrenceTest(TopydoTest):
 
         self.assertEqual(new_todo.due_date(), new_due)
 
+    def test_strict_recurrence3(self):
+        """
+        Strict recurrence where start date is in the past and there is no due
+        date, using + notation in expression.
+        """
+        past = date.today() - timedelta(8)
+        new_start = date.today() - timedelta(1)
+
+        self.stricttodo.set_tag(config().tag_start(), past.isoformat())
+        new_todo = advance_recurring_todo(self.stricttodo, p_strict=True)
+
+        self.assertEqual(new_todo.start_date(), new_start)
+
     def test_no_recurrence(self):
         self.todo.remove_tag('rec')
         self.assertRaises(NoRecurrenceException, advance_recurring_todo,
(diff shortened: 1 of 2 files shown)


davesteele:
Have you tested both forms of recurrence, normal and strict? I've found I need to add the due date to get one of them to work right (don't remember which right now). Dropping the due date could mess with that.

fuzzdk:
That is a good point!

I've now tested it with strict occurrence:

I have made this original task that I'm going to mark as done:
2024-07-22 Test rec:+1m  t:2024-07-20

(test run on 2024-07-22)

Wanted output:
2024-07-22 Test rec:+1m  t:2024-08-20

With changes in PR:
2024-07-22 Test rec:+1m  t:2024-08-22

Without changes in this PR:
2024-07-22 Test rec:+1m  t:2024-08-22 due:2024-08-22

So both wrong before and after. I'll try to extend the testcase and see if I can fix this.

bram85:
Maybe a bit childish question: why would topydo have to follow SimpleTask and not the other way around?

I'd be more in favor of having things explicit, i.e. have an explicit due date in next occurrences rather than implicit ones.

davesteele:
You've forced me to look harder at my concern here. It turns out it involves a personal preference, and not directly related to where the PR is aimed. I still think it points to an issue to be addressed.

I don't tend to work with either creation or due dates, and I believe both tools should accommodate that. It seems obvious to me that normal/default recurrence should be based on the completion date (i.e. now), and that strict recurrence should be based on only the threshold date. If a due date is present, it should be pushed the same amount as the threshold date. Due dates should be added to the new task only if there is one on the previous.

Simpletask says that strict recurrence is based on "the original due or threshold date". Topydo uses the due date (per the wiki) and apparently also the creation date (per the example above). **I think both are wrong**, and that only the threshold date should be used. I may be missing something here - I don't get why one would do it differently.

My specific issue with Topydo strict recurrence is that it breaks if neither the creation or due date is defined (I'm out on a limb a bit here - it's been a while and I didn't explicitly test this today). SimpleTask works in this case.

This PR steps into this issue by dropping due dates in some cases (my current personal workaround involves matching threshold and due dates). I would need to make sure that any changes here have don't mess with my task flow. 

I'm fine with explicit, but I agree with the message I'm hearing from @fuzzdk - don't add a due date based on a task for which I chose to exclude it. Additional, recurrence should work with only a threshold date available.

fuzzdk:
I think it is important to take into account the 4 different possibilities in the input of whether dates are present or not

### Has due date and threshold date

This case is simple. Both increment the due date and threshold date.
Example:
```
pay rent due:2024-08-01 t:2024-07-20 rec:+1m
```
here it is simple with strict recurrence: just increment both dates. 

### Only has due date
Example:
```
pay rent due:2024-08-01 rec:+1m
```
here the output should increment the due date and not set a threshold date - as it would be impossible to guess the threshold date.

### Only has threshold date
Example:
```
Watch the next episode of a specific tv-series t:2024-07-24 rec:+7d
```
If the episode of my favorite tv-series gets available for streaming on Wednesdays, this could be a reasonable task to have as a reminder. There is no due date - not even an implicit one, as it doesn't matter which day it is watched. The only way to say there is no due date is to not have one. So the recurring task that is inserted when I mark it as done should not have one. This and only this case out of the 4 is what this PR is about.

### Has no dates
Example:
```
Check oil on car rec:1m
```
Here it gets into a matter of taste! What does it actually mean? I wouldn't expect different programs to understand this corner case the same way.

fuzzdk:
I have now also fixed the strict recurrence in the case where there is a start date. I've created unit tests for the changes. I've manually tested the 8 possibilities of strict/normal recurrence, due date/no due date and start date/no start date which behaves as it should. If I can do anything more for this PR please let me know.

davesteele:
That is great. I will look at it later today.

davesteele:
This is better. Thank you.

PR #336, opened 2026-03-16
Remove unused requirement of pytest-runner

horrad:
Building the topydo package on OpenBSD we found the requirement of pytest-runner is not needed.

Even make test did succeed 100%.

If there is no need for this in other systems, it can be removed.


davesteele:
Thanks

PR #339, opened 2026-08-11
Todofile: replace codecs.open with open

hydrargyrum:
it seems open() supports encoding param since earlier than python 3.4 so we should be safe

### DIFF
--- a/topydo/lib/TodoFile.py
+++ b/topydo/lib/TodoFile.py
@@ -35,9 +35,8 @@ class TodoFile(object):
         """ Reads the todo.txt file and returns a list of todo items. """
         todos = []
         try:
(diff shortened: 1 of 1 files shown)


topydo/lib/TodoFile.py
```diff
@@ -51,14 +50,13 @@ def write(self, p_todos):
...
+        with open(self.path, 'w', encoding="utf-8") as todofile:
+            if isinstance(p_todos, list):
+                for todo in p_todos:
+                    todofile.write(str(todo))
+            else:
+                todofile.write(p_todos)
 
+            todofile.write("\n")
+            
         todofile.close()
```
hydrargyrum: this line should be removed as the `with` should take care of closing, let me repush without it

it was re-introduced by the conflict solve

